PR #4, opened 2025-12-21
Support turn and user message cancellation

xtmq:
Implementation of features to cancel user messages and turns

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -9,6 +9,7 @@ import type {
     SetDefaultModelParams,
     SetDefaultModelResponse
 } from "./app-server";
+import type {TurnCompletedNotification } from "./app-server/v2";
 import type {JsonValue} from "./app-server/serde_json/JsonValue";
 import type {Model} from "./app-server/v2";
 import {ModelId} from "./ModelId";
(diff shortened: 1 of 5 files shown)


src/CodexAcpServer.ts
```diff
@@ -153,25 +153,34 @@ export class CodexAcpServer implements acp.Agent {
...
+                return {
+                    stopReason: "cancelled",
+                };
             }
 
+            return {
+                stopReason: "end_turn",
+            };
+        } catch (err) {
+            console.error(`[CodexAcpServer] prompt() failed:`, err);
```
slapoguzov: how do we parse such errors in acp client? will they be displayed in our chat or silently ignored?

src/CodexAcpClient.ts
```diff
@@ -119,13 +124,27 @@ export class CodexAcpClient {
             model: null,
         });
 
-        await this.codexClient.awaitTurnCompleted();
+        // Save turnId if callback provided
+        if (onTurnStarted) {
```
slapoguzov: Instead, we can process the turn/started event in createUpdateEvent() and save the turnId to sessionState there. Then we don’t need onTurnStarted() here

src/CodexAcpServer.ts
xtmq: Ignored - this is exclusevly for logging

slapoguzov (approved):

PR #5, opened 2025-12-21
Fix file creation

slapoguzov:
Fix file creation

### DIFF
--- a/src/CodexEventHandler.ts
+++ b/src/CodexEventHandler.ts
@@ -153,19 +153,31 @@ export class CodexEventHandler {
     }
 
     private async createPatchContent(change: FileUpdateChange): Promise<ToolCallContent | null> {
-        const oldContent = change.kind.type === "add"
-            ? null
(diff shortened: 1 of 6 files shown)


src/CodexEventHandler.ts
```diff
@@ -153,19 +153,33 @@ export class CodexEventHandler {
...
+                oldText: null,
+                newText: change.diff,
+                path: change.path,
+            }
+        }
+
+        const oldContent = change.kind.type === "add" ? "" : await readFile(change.path, { encoding: "utf8" });
         const newContent = applyPatch(oldContent, change.diff);
-        if (!newContent) {
+        if (newContent === false) {
```
xtmq, line 168: we can not remove files using this event?

xtmq (approved):

src/CodexEventHandler.ts
slapoguzov: will handle it later

PR #7, opened 2025-12-21
LLM-22217 Support command approvals & edit approvals

slapoguzov:


src/CodexAcpClient.ts
```diff
@@ -85,7 +85,7 @@ export class CodexAcpClient {
             modelProvider: this.modelProvider,
             model: null,
             cwd: request.cwd,
-            approvalPolicy: "never",
+            approvalPolicy: "on-request",
```
xtmq, line 88: don't we have something like `brave-mode` in `request`?

src/CodexAcpClient.ts
```diff
@@ -104,10 +104,12 @@ export class CodexAcpClient {
...
 
     async sendPrompt(
-        request: acp.PromptRequest, 
-        eventHandler: (result: ServerNotification) => void
+        request: acp.PromptRequest,
+        eventHandler: (result: ServerNotification) => void,
+        approvalHandler: ApprovalHandler
     ): Promise<TurnCompletedNotification> {
         this.codexClient.onServerNotification(request.sessionId, eventHandler);
+        this.codexClient.onApprovalRequest(request.sessionId, approvalHandler);
```
xtmq, line 112: Shouldn't we cleanup these collections at some point? On session terminartion maybe?

xtmq (approved):

src/CodexAcpClient.ts
slapoguzov: yeah, it will be passed to `sandbox` paramter

slapoguzov: When @NikolaiSviridov implements support for modes

PR #10, opened 2025-12-26
Token Usage

slapoguzov:
- Add token_count to PromptResponse._meta.quota with per-turn token usage stats
- Add /run-codex skill for testing with real Codex during development

### DIFF
--- a/src/CodexAcpServer.ts
+++ b/src/CodexAcpServer.ts
@@ -8,11 +8,13 @@ import type {Model} from "./app-server/v2";
 import type {ReasoningEffort} from "./app-server";
 import {ModelId} from "./ModelId";
 import {AgentMode} from "./AgentMode";
+import type {TokenCount} from "./TokenCount";
 
 
 export interface SessionState {
     sessionMetadata: SessionMetadata;
     currentTurnId: string | null;
+    lastTokenUsage: TokenCount | null;
 }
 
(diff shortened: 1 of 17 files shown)


src/CodexAcpServer.ts
```diff
@@ -205,6 +211,14 @@ export class CodexAcpServer implements acp.Agent {
         }
     }
 
+    private buildQuotaMeta(sessionState: SessionState): { quota: { token_count: TokenCount | null } } {
```
xtmq, line 214: Now we pass codex specific token data class to meta, don''t we have a proper data model for quota in ACP?

slapoguzov: > Now we pass codex specific token data class to meta

TokenCount is our own class. I created it specifically to avoid passing Codex classes into ACP and to make it easier to track changes in Codex event structures.

>Don''t we have a proper data model for quota in ACP?

yes, it would be nice. I will talk with Anna about this but for now it's better to have something to be able to track this in the evals

slapoguzov: @anna239 what do you think about this? is it ok right now to pass token usage in meta? Or it's better to wait a proper support from ACP?

anna239: @slapoguzov sure, it's ok to pass in meta for now

xtmq (approved):

PR #13, opened 2026-01-05
LLM-21798 JetBrains AI authentication method

dtim:


src/CodexAcpClient.ts
```diff
@@ -75,6 +111,10 @@ export class CodexAcpClient {
     }
 
     async authRequired(): Promise<Boolean> {
+        if (this.skipAuthCheck) {
+            return false;
+        }
+
         const response = await this.codexClient.accountRead({refreshToken: false})
         return response.requiresOpenaiAuth && !response.account;
```
AlexandrSuhinin, line 127: requiresOpenaiAuth must return false if custom provider is configured, so skipAuthCheck is redundant

src/__tests__/CodexACPAgent/CodexAcpClient.test.ts
```diff
@@ -59,6 +59,32 @@ describe('ACP server test', { timeout: 40_000 }, () => {
         await expect(transportDump).toMatchFileSnapshot("data/auth-with-key.json");
     });
 
+    it('should authenticate with JetBrains AI', async () => {
+        const codexAcpAgent = fixture.getCodexAcpAgent();
+
+        await codexAcpAgent.initialize({protocolVersion: 1});
+        await fixture.getCodexAcpClient().logout();
+        fixture.clearCodexConnectionDump();
```
AlexandrSuhinin: fixture.clearCodexConnectionDump() is used to  partially clear connection history so it easier to compare it.
Redundant here - message history is not checked.

AlexandrSuhinin (approved):

src/CodexAcpClient.ts
dtim: The custom provider is passed (via the app-server protocol) only during the call to `thread/start`, and the `this.codexAcpClient.authRequired()` call in `CodexAcpServer.newSession` is performed before `thread/start`.

And we can't just remove the `authRequired` check: the ACP client tries to create a new session, catches the "auth required" error and triggers the authentication process. If we remove the check, the `authenticate` method will never be called.

So `skipAuthCheck` is a hacky way to enable and disable auth check (it is set to `true` only when `jetbrains-ai` auth is requested – and it does not actually call Codex at this point, it only sets a new config, – and reset it to `false` for any other auth method).

But I will look at the better way to do the trick (I agree, the `skipAuthCheck` is unclear), thank you!

PR #17, opened 2026-01-12
feat: add default model meta

NikolaiSviridov:


src/CodexAcpServer.ts
```diff
@@ -168,6 +168,9 @@ export class CodexAcpServer implements acp.Agent {
                 modelId: ModelId.fromComponents(model, effort.reasoningEffort).toString(),
                 name: `${model.displayName} (${effort.reasoningEffort})`,
                 description: `${model.description} ${effort.description}`,
+                _meta: {
```
AlexandrSuhinin, line 171: available models returned from newSession. And here we already have "modelId" which can be treated as default.
We should avoid extending protocol when it's possible.

AlexandrSuhinin (changes requested):

PR #21, opened 2026-01-15
Add logging for auth flow

xtmq:


src/CodexAcpServer.ts
```diff
@@ -112,15 +113,24 @@ export class CodexAcpServer implements acp.Agent {
     async newSession(
         _params: acp.NewSessionRequest,
     ): Promise<acp.NewSessionResponse> {
-        if (await this.runWithProcessCheck(() => this.codexAcpClient.authRequired())) {
+        logger.log("Starting new session...");
```
slapoguzov, line 131: why do we need "..."?

slapoguzov (approved):

PR #22, opened 2026-01-16
LLM-22554 Fix incorrectly handled file deletion

KirillAlekhnovich:
Send patch event kind in the meta field so client can set mandatory newText field to null

### DIFF
--- a/src/__tests__/CodexACPAgent/data/file-change-delete-raw-content.json
+++ b/src/__tests__/CodexACPAgent/data/file-change-delete-raw-content.json
@@ -5,18 +5,18 @@
       "sessionId": "test-session-id",
       "update": {
         "sessionUpdate": "tool_call",
-        "toolCallId": "file-change-raw",
+        "toolCallId": "file-delete-raw",
         "title": "Editing files",
         "kind": "edit",
         "status": "completed",
         "content": [
           {
             "type": "diff",
-            "oldText": null,
+            "oldText": "fun main() {\n    println(\"Hello, World!\")\n}\n",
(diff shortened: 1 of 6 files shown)


src/__tests__/CodexACPAgent/data/file-change-delete-raw-content.json
```diff
@@ -0,0 +1,26 @@
...
+        "sessionUpdate": "tool_call",
+        "toolCallId": "file-delete-raw",
+        "title": "Editing files",
+        "kind": "edit",
+        "status": "completed",
+        "content": [
+          {
+            "type": "diff",
+            "oldText": "fun main() {\n    println(\"Hello, World!\")\n}\n",
+            "newText": "fun main() {\n    println(\"Hello, World!\")\n}\n",
```
xtmq, line 16: why newText is not null in that case?
I would say: 
oldText != null, newText != null -> change
oldText == null -> add
newText == null -> remove

xtmq (changes requested):

src/__tests__/CodexACPAgent/data/file-change-delete-raw-content.json
KirillAlekhnovich: That's exactly the way we treat file changes in A2UX, but unfortunately `newText` in current ACP version is not nullable. See https://agentclientprotocol.com/protocol/tool-calls#diffs 
Due to that, I'm passing the patch modification kind in the `_meta` field and setting `newText` to `null` on the client side.

xtmq (approved):

PR #27, opened 2026-01-22
kill codex process if it doesn't exit naturally

AlexandrSuhinin:


xtmq (approved):

src/index.ts
```diff
@@ -50,6 +50,12 @@ function startAcpServer() {
     const codexConnection = startCodexConnection(codexPath);
     process.stdin.on("close", (chunk: Buffer) => {
         codexConnection.process.stdin.end();
+        // Kill the codex process if it doesn't exit naturally
+        setTimeout(() => {
```
xtmq, line 54: I would add more logging to the whole termination process to understanbd later where we fail to terminate correctly

slapoguzov (approved):

PR #28, opened 2026-01-30
MCP support and refactorings

AlexandrSuhinin:
- **extract duplicated function**
- **remove redundant mode parameter on session creation**
- **remove SessionMetadata from SessionState**
- **remove redundant models state fields**
- **clarify name**
- **remove duplication inside resume and newSession**
- **log: remove unnecessary data escaping**
- **update supported codex version & regenerate types**
- **fix typecheck errors and expected communication snapshots**
- **fix default version in clientInfo**
- **LLM-23643: support CPUs from before 2013**
- **log codex process termination**
- **use mcp servers provided via ACP**


src/CodexAcpClient.ts
```diff
@@ -235,7 +235,7 @@ export class CodexAcpClient {
         return await this.codexClient.awaitTurnCompleted();
     }
 
-    async setModel(params: SetDefaultModelParams): Promise<SetDefaultModelResponse> {
+    async setDefaultModel(params: SetDefaultModelParams): Promise<SetDefaultModelResponse> {
```
slapoguzov, line 238: is it unused?

slapoguzov (approved):

src/CodexAcpClient.ts
AlexandrSuhinin: yes, leftovers from initial implementation when we changed the default model in the global config.
Now we set model via app-sever API on prompt sent.

PR #37, opened 2026-02-10
load session support

nizienko:
added session list and load session methods support

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -14,6 +14,8 @@ import type {
 import type {JsonValue} from "./app-server/serde_json/JsonValue";
 import {ModelId} from "./ModelId";
 import {AgentMode} from "./AgentMode";
+import path from "node:path";
+import {logger} from "./Logger";
 import type {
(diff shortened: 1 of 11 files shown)


src/CodexAcpClient.ts
```diff
@@ -290,6 +319,105 @@ export class CodexAcpClient {
...
+            if (path.isAbsolute(requestedCwd)) {
+                return thread.cwd === requestedCwd;
+            }
+            const requestedBase = path.basename(requestedCwd);
+            return path.basename(thread.cwd) === requestedBase;
+        };
+
+        const preferredProvider = this.getModelProvider();
+        const modelProviders = preferredProvider ? [preferredProvider] : [];
+        const [listResponse, loadedResponse] = await Promise.all([
```
NikolaiSviridov: Why are using 1 cursor for 2 endpoints and then using `listResponse.nextCursor ?? loadedResponse.nextCursor ?? null` in return? It seems incorrect, maybe I'm missing something

NikolaiSviridov (approved):

PR #51, opened 2026-03-03
Update update guidelines and stabilize tests

ishulgin:
- Force file-based credential storage via temp CODEX_HOME in auth test. Otherwise, they fail when running by sandboxed codex.
- Stabilize snapshot comparison by replacing response payload with a placeholder.
- Update AGENTS.md test/style guidance for explicit switch cases and snapshot placeholders.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,6 +12,8 @@
 
 - Keep edits consistent with existing formatting.
 - When adding env/config knobs, document them in `readme-dev.md`.
+- When updating discriminated-union/event `switch` statements, do not add a trailing fallback like `return null` only to satisfy TypeScript.
+- Handle each variant with an explicit `case`; if intentionally ignored, use an explicit no-op case.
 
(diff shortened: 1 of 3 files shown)


src/__tests__/CodexACPAgent/data/auth-with-key.json
```diff
@@ -87,89 +87,7 @@
 }
 {
   "eventType": "response",
-  "data": [
```
slapoguzov, line 90: I think that the list of models is important for us, we want to track changes in it. Why is this unstable? is it about all fields? Or the order?

slapoguzov (changes requested):

src/__tests__/CodexACPAgent/data/auth-with-key.json
ishulgin: Reverted this change

slapoguzov (approved):

PR #52, opened 2026-03-04
LLM-24890 Generate a chat name based on the first message

NikolaiSviridov:


src/CodexEventHandler.ts
```diff
@@ -130,19 +118,36 @@ export class CodexEventHandler {
...
-            case "model/rerouted":
-                return this.createModelReroutedEvent(notification.params);
-            case "fuzzyFileSearch/sessionUpdated":
-                return this.handleFuzzyFileSearchSessionUpdated(notification.params);
-            case "fuzzyFileSearch/sessionCompleted":
-                return this.handleFuzzyFileSearchSessionCompleted(notification.params);
         }
     }
 
+    private extractNotificationThreadId(notification: ServerNotification): string | null {
```
AlexandrSuhinin, line 130: Why we need this unsafe method?
App-server listener should already be session-scoped.
Plus we can take threadId directly if we pass specific type after switch-case.

src/CodexAcpServer.ts
```diff
@@ -40,6 +39,7 @@ export interface SessionState {
     rateLimits: RateLimitsMap | null;
     account: Account | null;
     cwd: string;
+    threadRenameState: "notScheduled" | "running" | "done";
```
AlexandrSuhinin, line 42: Can we simplify the state here? Ideally we don't save it and on first prompt check if it's a first user prompt and schedule rename if it is.
In worst case we can introduce this flag instead of rename state.

src/CodexAcpServer.ts
```diff
@@ -679,6 +689,88 @@ export class CodexAcpServer implements acp.Agent {
         }
     }
 
+    private scheduleThreadRenameFromContext(
+        sessionState: SessionState,
+        modelId: ModelId,
+        summary: "none" | null,
```
AlexandrSuhinin, line 695: Maybe we can always disable streaming reasoning for thread naming process?
Do they stream this reasoning in CLI?

ModelId is already defined in the sessionState, why we need to duplicate?

src/CodexAppServerClient.ts
```diff
@@ -174,13 +211,15 @@ export class CodexAppServerClient {
...
     }
 
-    //TODO create type-safe helper
-    async awaitTurnCompleted(): Promise<TurnCompletedNotification> {
-        return await new Promise((resolve) => {
-            this.connection.onNotification("turn/completed", (event: TurnCompletedNotification) => {
-                resolve(event);
-            });
-        });
+    async awaitTurnCompletedForThread(threadId: string, turnId: string | null = null): Promise<TurnCompletedNotification> {
```
AlexandrSuhinin, line 214: Do you really need to wait for specific turn? You can't have multiple running turns in the same session.
Also to me it's enough to add threadId as parameter, no need to duplicate it in the name.
awaitTurnCompleted(threadId)

src/CodexAppServerClient.ts
```diff
@@ -216,6 +255,73 @@ export class CodexAppServerClient {
         }
     }
 
+    private handleTurnCompleted(event: TurnCompletedNotification): void {
```
AlexandrSuhinin, line 258: Why we need all these methods?

src/__tests__/CodexACPAgent/CodexAcpClient.test.ts
```diff
@@ -265,9 +264,9 @@ describe('ACP server test', { timeout: 40_000 }, () => {
 
         // Trigger notifications after both prompts - should produce only 3 events, not 6
         const serverNotifications: ServerNotification[] = [
```
AlexandrSuhinin, line 267: Why we renamed it and introduced parameter for the function loadNotifications? As far as I can see it only have single usage.

src/CodexAcpClient.ts
```diff
@@ -464,6 +462,96 @@ export class CodexAcpClient {
...
+                cwd: null,
+                approvalPolicy: "never",
+                sandbox: "read-only",
+                config: null,
+                baseInstructions: null,
+                developerInstructions: null,
+            });
+            forkedThreadId = fork.thread.id;
+
+            const started = await this.codexClient.turnStart({
```
AlexandrSuhinin, line 488: Is this approach copied from VS Code implementation including prompt text?
I would expect official API from the app-server, we can ask about this codex team.

AlexandrSuhinin (changes requested):

src/CodexAcpClient.ts
```diff
@@ -373,9 +371,9 @@ export class CodexAcpClient {
             model: modelId.model,
         });
 
-        // Wait for turn completion
-        // If turnInterrupt() was called, Codex will send turn/completed event with status "interrupted"
-        return await this.codexClient.awaitTurnCompleted();
+        // Wait for completion of the specific started turn.
+        // If turnInterrupt() was called, Codex will send turn/completed with status "interrupted".
+        return await this.codexClient.awaitTurnCompletedForThread(request.sessionId, started?.turn?.id ?? null);
```
xtmq, line 376: it seems it is a valid fix and it should be merged even withpout the feature itself. but worth an additional test which reproduces possible issue with previous versions without specific thread id

src/CodexAcpClient.ts
xtmq: It is possible an ACP agent does not support thread name providing. For that case we can fallback to some AIA logic which will be written and tested inside AIA chat and will be common for all agents. 

I also don't think we hgave to write such business logic in the adapter with custom promts, models, params, etc

PR #79, opened 2026-03-25
LLM-26016 [Codex] Show mcp initialization errors in chat

NikolaiSviridov:


src/CodexAcpServer.ts
```diff
@@ -700,6 +705,52 @@ export class CodexAcpServer implements acp.Agent {
         };
     }
 
+    private async publishMcpStartupStatus(sessionId: string, mcpStartup: McpStartupCompleteEvent): Promise<void> {
```
AlexandrSuhinin: Here we just map startup event into session update event. I believe that we have CodexEventHandler for this.

src/__tests__/CodexACPAgent/CodexAcpClient.test.ts
```diff
@@ -135,6 +135,48 @@ describe('ACP server test', { timeout: 40_000 }, () => {
         expect(newSessionResponse.sessionId).toBeDefined()
     })
 
+    it('should forward MCP startup failures to ACP session updates', async () => {
```
AlexandrSuhinin: Since we just added a mapping for codex to acp event, I would rather expect simple mapping test as we did for other CodexEventHandler logic.

PR #84, opened 2026-04-01
Use codex from npm dependency & prepare for npm publish

AlexandrSuhinin:


package.json
```diff
@@ -28,17 +39,18 @@
...
+    "diff": "^8.0.3",
+    "esbuild": "^0.25.0",
     "@types/node": "^24.10.1",
     "mcp-hello-world": "^1.1.2",
     "tsx": "^4.20.6",
     "typescript": "^5.9.3",
     "vitest": "^4.0.10"
   },
   "dependencies": {
-    "@agentclientprotocol/sdk": "^0.12.0",
```
xtmq, line 41: I guess it also should be regular dependency?

package.json
```diff
@@ -1,9 +1,20 @@
 {
-  "name": "codex-acp",
+  "name": "@jetbrains/codex-acp",
```
xtmq, line 2: it will be in acp repo, probably we have to use their naming scheme

xtmq (approved):

PR #90, opened 2026-04-09
LLM-26222 Show input params, result of running mcp tools, LLM-26016 Show mcp initialization errors in chat, LLM-26758 Failed to initialize ACP process

NikolaiSviridov:
United all the issues in one MR because they are all related to MCP, the latest update affects all of them, and it was easier to push them together at once

- LLM-26222 Show input params, result of running mcp tools

- LLM-26016 Show mcp initialization errors in chat
Versions were added so ACP can match an MCP startup-complete event to the correct session startup window, instead of showing initialization errors from some earlier or unrelated startup state

- LLM-26758 Failed to initialize ACP process
Changes related to `const codexPath = resolveCodexPath();`.
Added `scripts/build-macos-local.sh`, which is used for local testing of the binary. The current MCP plugin settings do not allow ACP to use them, so I tested it via the binary for integrated Codex. That is how I found that it is broken on the current `main` branch.


scripts/build-macos-local.sh
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
slapoguzov: you can use scripts from packages.json instead

src/CodexAcpServer.ts
```diff
@@ -41,6 +49,11 @@ export interface SessionState {
     account: Account | null;
     cwd: string;
     sessionMcpServers?: Array<string>;
+    mcpToolLogs: Map<string, Array<string>>;
```
slapoguzov: we shouldn't store the entire output in memory, let's only send delta

src/CodexAcpClient.ts
```diff
@@ -275,6 +277,22 @@ export class CodexAcpClient {
         return startup.ready;
     }
 
+    async awaitMcpStartupResult(mcpStartupVersion: number): Promise<McpStartupCompleteEvent> {
+        return await this.codexClient.awaitMcpStartup(mcpStartupVersion);
+    }
+
+    onMcpServerStatusUpdated(handler: (event: McpServerStatusUpdatedNotification) => void): void {
```
slapoguzov: are this and two methods below used anywhere?

slapoguzov: If it is not, we also don't need mcpServerStatusVersion, mcpServerStatusUpdatedHandlers and mcpServerStatusHistory in CodexAppServerClient.ts amd code there can be significantly simplified

src/CodexAppServerClient.ts
```diff
@@ -76,8 +83,10 @@ export class CodexAppServerClient {
                 }
                 return;
             }
-
             const serverNotification = data as ServerNotification;
+            if (serverNotification.method === "mcpServer/startupStatus/updated") {
```
slapoguzov: if get rid of of mcp* fields, we can move this into one of handler

slapoguzov (changes requested):

src/CodexAcpServer.ts
```diff
@@ -359,6 +381,14 @@ export class CodexAcpServer implements acp.Agent {
         };
         this.sessions.set(sessionId, sessionState);
 
+        const requestedMcpServers = request.mcpServers ?? [];
+        if (requestedMcpServers.length > 0) {
+            this.pendingMcpStartupSessions.set(sessionId, {
+                requestedServers: new Set(requestedMcpServers.map(server => server.name)),
+            });
+            this.publishMcpStartupStatusAsync(sessionId, mcpStartupVersion);
```
slapoguzov, line 389: publishMcpStartupStatusAsync() would be clearer if it followed the same pattern as publishAvailableCommandsAsync(): keep the wrapper fire-and-forget, and move the async body into a separate async method with internal error handling. Here it matters even more because this path mutates session state and pendingMcpStartupSessions, so the inner method should also do cleanup in finally

slapoguzov (changes requested):

slapoguzov (approved):

PR #92, opened 2026-04-13
Implement session usage updates

egor-baranov:
Implement https://agentclientprotocol.com/rfds/session-usage RFD that is already present in SDKs


src/TokenCount.ts
xtmq, line 14: Why do we need TokenCount  and Usage structures for basically the same information?

src/CodexAcpServer.ts
```diff
@@ -751,6 +753,7 @@ export class CodexAcpServer implements acp.Agent {
                 });
                 return {
                     stopReason: "cancelled",
+                    usage: this.buildPromptUsage(sessionState.lastTokenUsage),
                     _meta: this.buildQuotaMeta(sessionState),
```
xtmq, line 757: shouldn't we now remove usage info from meta? cc @dtim

xtmq (approved):

src/CodexAcpServer.ts
egor-baranov: We now expose token usage through PromptResponse.usage, but _meta.quota is still consumed on the IDE side for quota accounting and telemetry (LlmResponseMetadataEvent). I think we can refactor it after migration to the new entities (in a different mr)

src/TokenCount.ts
egor-baranov: I think we can migrate to Usage from internal TokenCount but I'm not sure we should do it in a current mr

PR #100, opened 2026-04-20
LLM-27035 [Codex] Add missing elicitation, LLM-27036 [Codex] Improve mcp block detection

NikolaiSviridov:


i1bro:
Can you also add some marker in _meta that the permission request is for an MCP tool?
A boolean would be enough

src/CodexEventHandler.ts
```diff
@@ -193,6 +196,7 @@ export class CodexEventHandler {
             case "commandExecution":
                 return await createCommandExecutionUpdate(event.item);
             case "mcpToolCall":
+                this.pendingMcpApprovals?.record(event.threadId, event.item.server, event.item.id);
```
slapoguzov: record() fires on every mcpToolCall start, pop() only on approvals, so auto-approved calls leave stale entries. Can we also pop() in the mcpToolCall branch of completeItemEvent to symmetrically clean up?

src/CodexAcpServer.ts
```diff
@@ -705,11 +707,14 @@ export class CodexAcpServer implements acp.Agent {
         sessionState.lastTokenUsage = null;
 
         try {
-            const eventHandler = new CodexEventHandler(this.connection, sessionState);
+            const pendingMcpApprovals = new PendingMcpApprovals();
```
slapoguzov: we should cleanup this on we on serverRequest/resolved for cases when a turn is interrupted mid-approval

slapoguzov: Also, PendingMcpApprovals is shared between CodexEventHandler and CodexElicitationHandler. Ownership is unclear, both constructors take it as an optional param, and the record/pop lifecycle is split across two files, making it easy to drift out of sync. I suggest moving PendingMcpApprovals (also you can inline this class) into CodexElicitationHandler as a private field and add a handleNotification that will handle mcp events

slapoguzov (changes requested):

slapoguzov (approved):

PR #102, opened 2026-04-20
Fix/provider name

i1bro:
## Summary
- pass the selected gateway `providerName` into the Codex config/auth flow
- allow `providerName` in the gateway auth request metadata type
- keep the change scoped to provider-name plumbing only

## Why
Some gateway auth flows need to know which provider was selected. This branch preserves that value end-to-end and removes the related TypeScript errors.

## Testing
- `npm test`


slapoguzov (approved):

AlexandrSuhinin:
Why have we merged this change? Do we plan to support "names" in all bundled agents?
I don't think that acp should know something about client choice, it should only provide a way to configure baseurl + custom headers. 
@slapoguzov

PR #103, opened 2026-04-20
Fix/edit tool calls

i1bro:
## Summary
- capture file-change context from app-server events for later approval and tool-call updates
- use edited file paths as edit tool-call titles instead of the generic "Editing files"
- attach file locations and diff data to edit tool calls, with a turn-diff fallback when needed
- serialize session updates so the initial tool call is emitted before its completion update

## Why
This makes edit tool calls understandable in the UI and avoids races where a completion update can arrive before the corresponding tool call exists.

## Testing
- `npm test`


src/CodexAppServerClient.ts
```diff
@@ -190,22 +193,56 @@ export class CodexAppServerClient {
      * Registers a notification handler for a specific session.
      * Replaces any existing handler for the same session, preventing handler accumulation.
      */
-    onServerNotification(sessionId: string, callback: (event: ServerNotification) => void) {
+    onServerNotification(sessionId: string, callback: (event: ServerNotification) => void | Promise<void>) {
```
AlexandrSuhinin: Would be more clean to expect only Promise<void> here and avoid double types.

src/CodexAppServerClient.ts
```diff
@@ -60,6 +61,8 @@ const FileChangeApprovalRequest = new RequestType<
 export class CodexAppServerClient {
     readonly connection: MessageConnection;
     private approvalHandlers = new Map<string, ApprovalHandler>();
+    private readonly notificationHandlers = new Map<string, (event: ServerNotification) => void | Promise<void>>();
```
AlexandrSuhinin: This file is meant to be thin jsonrpc layer around codex types that would simplify jsonrpc notification subscription and would declare missing function signatures.
It's already quite overcomplicated, I would try to avoid adding new entities and fields here.
Can we express notification handlers/queues as a wrapper over notification subscription and move it to CodexAcpClient?

src/CodexAcpClient.ts
```diff
@@ -359,7 +359,7 @@ export class CodexAcpClient {
 
     async subscribeToSessionEvents(
         sessionId: string,
-        eventHandler: (result: ServerNotification) => void,
+        eventHandler: (result: ServerNotification) => void | Promise<void>,
```
AlexandrSuhinin: let's remain only Promise<void> for clarity?

src/CodexAppServerClient.ts
```diff
@@ -190,22 +193,56 @@ export class CodexAppServerClient {
...
+                    });
+                    this.notificationQueues.set(sessionId, this.trackNotificationQueue(sessionId, next));
+                }
+            } catch (error) {
+                logger.error("Error handling server notification", error);
+            }
         }
     }
 
+    async flushServerNotifications(sessionId: string): Promise<void> {
```
AlexandrSuhinin: I'm not sure why do we need a queue and then why we need to flush it instead of handling events?

src/CodexApprovalHandler.ts
```diff
@@ -20,13 +28,16 @@ const APPROVAL_OPTIONS: acp.PermissionOption[] = [
 export class CodexApprovalHandler implements ApprovalHandler {
     private readonly connection: acp.AgentSideConnection;
     private readonly sessionState: SessionState;
+    private readonly approvalContext: ApprovalContextStore;
 
     constructor(
         connection: acp.AgentSideConnection,
-        sessionState: SessionState
+        sessionState: SessionState,
+        approvalContext: ApprovalContextStore,
```
AlexandrSuhinin: Maybe it's better to just inline ApprovalContextStore fields here and remove interface?
Looks like permission logic is scattered between CodexApprovalHandler and EventHandler. I would either separate them clearly and move permission-related code into CodexApprovalHandler or merge these two classes into one.
I would avoid having shared state between these two classes.

src/CodexAppServerClient.ts
i1bro: Because sendPrompt() must not return until all earlier events have been handled.

src/CodexAcpServer.ts
```diff
@@ -706,12 +704,12 @@ export class CodexAcpServer implements acp.Agent {
...
-            const approvalContext = createApprovalContextStore();
-            const eventHandler = new CodexEventHandler(this.connection, sessionState, approvalContext);
-            const approvalHandler = new CodexApprovalHandler(this.connection, sessionState, approvalContext);
-            await this.codexAcpClient.subscribeToSessionEvents(params.sessionId,
+            const eventHandler = new CodexEventHandler(this.connection, sessionState);
+            this.codexAcpClient.subscribeToSessionEvents(
+                params.sessionId,
                 (event) => eventHandler.handleNotification(event),
-                approvalHandler);
+                eventHandler
```
AlexandrSuhinin: Why we need to pass eventHandler twice?

src/CodexEventHandler.ts
```diff
@@ -1,70 +1,113 @@
...
 
-export { stripShellPrefix };
+const APPROVAL_OPTIONS: acp.PermissionOption[] = [
+    { optionId: "allow_once", name: "Allow Once", kind: "allow_once" },
+    { optionId: "allow_always", name: "Allow for Session", kind: "allow_always" },
+    { optionId: "reject_once", name: "Reject", kind: "reject_once" },
+];
 
-export class CodexEventHandler {
+export class CodexEventHandler implements ApprovalHandler {
```
AlexandrSuhinin: Do we really need the interface then?

src/CodexAcpClient.ts
```diff
@@ -584,6 +608,15 @@ export class CodexAcpClient {
 
 export type JsonObject = { [key in string]?: JsonValue }
 
+type SessionEventHandler = ApprovalHandler & {
```
AlexandrSuhinin: If we merge ApprovalHandler with event handler, then this type is redundant probably?

src/CodexAcpClient.ts
```diff
@@ -44,6 +44,7 @@ export class CodexAcpClient {
     private gatewayConfig: GatewayConfig | null;
     private pendingLoginCompleted: Promise<AccountLoginCompletedNotification> | null = null;
     private pendingAccountUpdated: Promise<AccountUpdatedNotification> | null = null;
+    private readonly sessionNotificationStates = new Map<string, SessionNotificationState>();
```
AlexandrSuhinin: I don't get why we need notification state to handle events. Can you explain?
I'd expect API similar to "appServer.subscribe(sessionId, handler)". When appServer just passes events to the handler.

src/CodexToolCallMapper.ts
```diff
@@ -472,3 +472,60 @@ function patchToDeletedContent(unifiedDiff: string): string | null {
...
+        oldText: change.kind.type === "add" ? null : diffContent.oldText,
+        newText: change.kind.type === "delete" ? "" : diffContent.newText,
+        path: change.path,
+        _meta: {
+            kind: change.kind.type,
+        },
+    };
+}
+
+function patchToDiffContent(unifiedDiff: string): { oldText: string; newText: string } | null {
```
AlexandrSuhinin: There is a `diff` dependency that is used to apply patches. Let's avoid writing diff parser.
See `createPatchContent` function.

src/CodexAcpClient.ts
i1bro: Changed to subscribeSession(sessionId, handler). The only remaining per-session state is a pending promise chain in CodexAppServerClient that serializes async handler delivery, so later events cannot overtake earlier ones.

src/CodexAcpClient.ts
```diff
@@ -397,7 +382,13 @@ export class CodexAcpClient {
 
         // Wait for turn completion
         // If turnInterrupt() was called, Codex will send turn/completed event with status "interrupted"
-        return await this.codexClient.awaitTurnCompleted();
+        const turnCompleted = await this.codexClient.awaitTurnCompleted();
+        await this.codexClient.awaitSessionIdle(request.sessionId);
+        return turnCompleted;
+    }
+
+    subscribeSession(sessionId: string, handler: Parameters<CodexAppServerClient["subscribeSession"]>[1]): void {
```
AlexandrSuhinin: Unclear signature for "API" - what is Parameters<CodexAppServerClient["subscribeSession"]>[1]?

src/CodexAppServerClient.ts
```diff
@@ -330,6 +351,21 @@ export type CodexConnectionEvent =
 
 type CodexRequest = DistributiveOmit<ClientRequest, "id">
 
+type SessionHandlerState = {
+    handler: SessionHandler;
+    pending: Promise<void>;
```
AlexandrSuhinin, line 356: Can you clarify why do we need state management to handle notifications instead of calling "handler" when event arrives?

src/CodexToolCallMapper.ts
```diff
@@ -256,94 +338,172 @@ function createSearchTitle(query: string | null, path: string | null): string {
...
+
+function formatPatchForOutput(patch: ReturnType<typeof parsePatch>[number]): string {
+    return formatPatch({
+        ...patch,
+        oldHeader: undefined,
+        newHeader: undefined,
+    }, FILE_HEADERS_ONLY);
+}
+
+function createDiffContentFromParsedPatch(
```
AlexandrSuhinin, line 446: Why we manually create patch lines? Don't we have it already in change?

src/CodexToolCallMapper.ts
```diff
@@ -256,94 +338,172 @@ function createSearchTitle(query: string | null, path: string | null): string {
...
+}
+
+function joinPatchLines(lines: string[]): string {
+    if (lines.length === 0) {
+        return "";
+    }
+    return lines.join("\n");
+}
+
+function normalizeDiffPath(fileName: string | undefined): string | undefined {
```
AlexandrSuhinin: filename?: string;
return type is not consistent - somewhere is string | undefined, somewhere type | null

AlexandrSuhinin (changes requested):

src/CodexAppServerClient.ts
i1bro: We need the per-session pending promise because notification handling is async: handleNotification() awaits createUpdateEvent() and session.update(), and file-change events may read files to reconstruct ACP diffs. The stored promise chain preserves event order and lets awaitSessionIdle() wait for all pending ACP updates before prompt completion.

src/CodexToolCallMapper.ts
i1bro: To convert the diff to ACP format and return as ACP "diff" content block. It needs oldText/newText

AlexandrSuhinin:
This PR is quite behind main branch. We agreed that @i1bro will collect issues that had to be solved by this and we'll fix them one by one.

PR #107, opened 2026-04-22
Use last-turn token usage for ACP context updates

egor-baranov:
Report ACP context usage from last-turn token usage instead of cumulative totals, and align the status context window output with the same source.


dtim (approved):
lgtm

PR #110, opened 2026-04-23
Advertise auth logout capability

benbrandt:
Route logout through unstable ACP auth support while keeping the
legacy authentication/logout extension method as a compatibility
shim.


benbrandt:
Depends on #109

AlexandrSuhinin (approved):

PR #113, opened 2026-04-23
Implement closeSession

benbrandt:
Probably needs a close look as it ended up needing to touch more than I expected


src/CodexAcpServer.ts
```diff
@@ -380,6 +407,7 @@ export class CodexAcpServer implements acp.Agent {
             supportedReasoningEfforts: currentModel?.supportedReasoningEfforts ?? [],
             supportedInputModalities: currentModel?.inputModalities ?? ["text", "image"],
             agentMode: AgentMode.getInitialAgentMode(),
+            pendingTurnId: null,
```
AlexandrSuhinin: Can single session have two active turns? Can we reuse currentTurnId here?

AlexandrSuhinin (changes requested):

AlexandrSuhinin:
Do we want to change the patch after https://github.com/agentclientprotocol/codex-acp/pull/123
 is merged?

benbrandt:
Yeah I can revisit after that is in

PR #114, opened 2026-04-23
Publish npm package with ACP scope

ishulgin:


ishulgin:
@benbrandt I updated scope to the `@agentclientprotocol`. Could you please help with configuring trusted publishing?

Also note that our existing workflow requires tags creation, and I would prefer to keep it like this for a while. However, would be nice to eventually unify CI approaches with claude-acp.

ishulgin:
I can assign only one reviewer here, so will just cc @AlexandrSuhinin

PR #115, opened 2026-04-24
Add more e2e tests

ishulgin:


AlexandrSuhinin:
I’m not sure we really need the mocked gateway and request assertions here.

I'd suggest implementing a simple ACP Client instead; we could use the API key and base URL to handle actual gateway authentication for a basic prompt. My thought is to prioritize E2E tests using that client—keeping mocks to a minimum so we can cover the longest execution paths first.

What do you think?

AlexandrSuhinin (approved):
But I'm not against merging these tests.

PR #123, opened 2026-04-27
Fix session pollution: route notifications by threadId

SteffenDE:
I found this issue when testing the new Codex ACP in Tidewave. Session updates from one session would also be sent to all other sessions. I asked Claude to fix this and it also found an issue with prompt results, but please double check that that's accurate.

notify() broadcasted every Codex notification to every registered session handler, so a stale handler from a previous prompt would re-emit the new session's deltas as ACP sessionUpdates on the old sessionId. awaitTurnCompleted() had a similar issue: it never disposed its turn/completed listener and didn't filter by threadId, so concurrent prompts across sessions could resolve to the wrong turn.

Filter thread-scoped notifications by params.threadId in notify(), and make awaitTurnCompleted thread-scoped with disposal. Non-thread events (account, mcp, app-list, warnings) still broadcast.


benbrandt:
My close session PR would be much simpler if we already address this first.

AlexandrSuhinin (approved):

benbrandt:
@AlexandrSuhinin I added some changes to account for tracking by turn id

AlexandrSuhinin (approved):

PR #125, opened 2026-04-29
Add more e2e tests

ishulgin:


AlexandrSuhinin:
I would also test permissions in different modes (edit permission is missed in Agent mode, exists for Read)

AlexandrSuhinin (approved):

PR #129, opened 2026-04-30
feat: LLM-25394 support Codex persist approval

NikolaiSviridov:


AlexandrSuhinin:
In general I'm not sure this is the best approach to handle such cases. I expect that we may want to have similar behavior for all agents, then probably we can come up with some common solution in the IDE. WDYT?

src/CodexAcpServer.ts
```diff
@@ -51,7 +51,7 @@ export interface SessionState {
     rateLimits: RateLimitsMap | null;
     account: Account | null;
     cwd: string;
-    sessionMcpServers?: Array<string>;
+    sessionMcpServers: Map<string, acp.McpServer>;
```
AlexandrSuhinin: Wouldn't change to map to simplify .get() in single place where performance issue is not expected.

src/CodexAcpServer.ts
```diff
@@ -372,7 +371,6 @@ export class CodexAcpServer implements acp.Agent {
 
         const accountResponse = await this.runWithProcessCheck(() => this.codexAcpClient.getAccount());
         const {sessionId, currentModelId, models, thread} = sessionMetadata;
-        const sessionMcpServers = this.resolveSessionMcpServers(requestedMcpServers, true);
```
AlexandrSuhinin, line 455: I'm not against, but is there a reason to mix these such refactorings with changes (inline constant)?

src/CodexAcpClient.ts
```diff
@@ -322,6 +323,24 @@ export class CodexAcpClient {
         return new Set(Object.keys(mcpServers));
     }
 
+    // Makes an ACP-provided MCP server durable so Codex core can append
+    // mcp_servers.<server>.tools.<tool>.approval_mode afterward:
+    // https://github.com/openai/codex/blob/main/codex-rs/core/src/mcp_tool_call.rs
+    async persistMcpServer(mcpServer: McpServer): Promise<void> {
```
AlexandrSuhinin: Do we also save mcp servers in global config?
How exactly this affect cli agent?

src/CodexCommands.ts
```diff
@@ -165,9 +165,7 @@ export class CodexCommands {
                     const resourceCount = (server.resources ?? []).length;
                     return `- ${server.name}: ${toolCount} tools, ${resourceCount} resources, auth=${server.authStatus}`;
                 });
-                const sessionServers = sessionState.sessionMcpServers
-                    ? sessionState.sessionMcpServers.map(serverName => `- ${serverName}`)
-                    : [];
+                const sessionServers = Array.from(sessionState.sessionMcpServers.keys()).map(serverName => `- ${serverName}`);
```
AlexandrSuhinin: Is this related to persist approval?

AlexandrSuhinin (changes requested):

src/CodexAcpClient.ts
NikolaiSviridov: removed that logic

src/CodexCommands.ts
NikolaiSviridov: it was related, but now not relevant

src/__tests__/CodexACPAgent/data/elicitation-tool-approval-all-persist.json
```diff
@@ -29,22 +29,26 @@
         {
           "optionId": "allow_once",
           "name": "Allow",
-          "kind": "allow_once"
+          "kind": "allow_once",
+          "_meta": "_meta"
```
AlexandrSuhinin: why _meta appeared?

src/CodexAcpServer.ts
```diff
@@ -56,7 +56,8 @@ export interface SessionState {
     cwd: string;
     fastModeEnabled: boolean;
     currentModelSupportsFast: boolean;
-    sessionMcpServers?: Array<string>;
+    sessionMcpServers: Array<acp.McpServer>;
+    configBackedMcpServerNames?: Set<string>;
```
AlexandrSuhinin, line 61: What if it's globally configured MCP?
Why this field is needed at all if you can get names from sessionMcpServers?

src/CodexAcpServer.ts
```diff
@@ -56,7 +56,8 @@ export interface SessionState {
     cwd: string;
     fastModeEnabled: boolean;
     currentModelSupportsFast: boolean;
-    sessionMcpServers?: Array<string>;
+    sessionMcpServers: Array<acp.McpServer>;
```
AlexandrSuhinin: Why we changed this to Array<acp.McpServer> if all usages reference .name only?

src/__tests__/CodexACPAgent/approval-events.test.ts
```diff
@@ -107,6 +107,74 @@ describe('Approval Events', () => {
             await promptPromise;
         });
 
+        it('should map execpolicy amendment approval to Codex decision', async () => {
```
AlexandrSuhinin, line 110: I would expect more explicit e2e test for persisted request.
E.g. start agent -> send prompt that should trigger permission -> always accept -> stop agent -> start new one & repeat -> check that permission request wasn't sent.

AlexandrSuhinin (changes requested):

src/CodexAcpServer.ts
```diff
@@ -120,7 +121,6 @@ export class CodexAcpServer implements acp.Agent {
                     list: { }
                 },
                 mcpCapabilities: {
-                    acp: false,
```
AlexandrSuhinin, line 129: Is this expected change?

AlexandrSuhinin (changes requested):

src/CodexAcpServer.ts
NikolaiSviridov: there was a strange typecheck error which I wanted to recheck before requesting a review. With rebase it's gone

AlexandrSuhinin:
Diverged too much from the main. Also solution is incomplete.
Believe that it's easier to reopen it in a new PR once we ready.

PR #133, opened 2026-05-04
feat: LLM-26878 Add “Fast” mode for Codex-agent

NikolaiSviridov:


src/ModelId.ts
```diff
@@ -1,42 +1,48 @@
...
 
-    static create(modelId: string, effort: ReasoningEffort): ModelId {
-        return new ModelId(modelId, effort);
+    static create(modelId: string, effort: ReasoningEffort, serviceTier: ServiceTier | null = null): ModelId {
+        return new ModelId(modelId, effort, serviceTier);
     }
 
     static fromString(modelId: string): ModelId {
-        const bracketMatch = modelId.match(/^(?<model>[^\[]+?)(?:\[(?<effort>[^\]]+)\])?$/);
+        const bracketMatch = modelId.match(/^(?<model>[^\[]+)\[(?<effort>[^\]]+)](?:@(?<serviceTier>.+))?$/);
```
AlexandrSuhinin: This is already too complicated. Especially we leak into the "protocol" codex-specific details.
I would rather expose this as a new model in the picker and would remap it internally into the same model with fast service. Though I'd avoid mapping based on the model id parsing.

AlexandrSuhinin (changes requested):

AlexandrSuhinin (approved):

NikolaiSviridov:
Dmitry Kozhevnikov approved

PR #143, opened 2026-05-09
Add /review command support

gabrielbergamini:
## Summary

This adds no-argument `/review` support to the ACP adapter using Codex app-server `review/start`.

The command starts a review of uncommitted changes in the current Codex session:

```text
/review
```

This intentionally does not add `/review <custom instructions>`, branch review, commit review, or detached review threads yet. Existing ACP built-in commands currently advertise `input: null`, so accepting ad hoc arguments for `/review` would create a mismatch between typed prompt input and command UI metadata. Custom review instructions can be added later once command input UX is clarified.

## What changed

- Added app-server client support for `review/start`.
- Added an ACP client helper that starts reviews in the current Codex thread.
- Published `/review` as a built-in ACP command.
- Mapped `/review` to `ReviewTarget: { type: "uncommittedChanges" }`.
- Rejected `/review <anything>` with a short usage message:

  ```text
  /review does not accept arguments yet.
  ```

- Tracked the active review turn ID so ACP cancellation and session close can interrupt an in-progress review.
- Rendered live review-mode events:
  - `enteredReviewMode` -> `Code review started: current changes`
  - `exitedReviewMode` -> review text followed by `Code review finished`

- Kept history replay formatting consistent with live review event rendering.
- Added tests for app-server review wiring, command behavior, active review cancellation, prompt-level cancellation/error propagation, and review event rendering.

Implementation note: the adapter passes `delivery: "inline"` to `review/start`, which is the app-server mode for running the review in the current thread. When `review/start` returns, its turn ID is stored as the session's current turn until the prompt completes, matching the lifecycle used by normal prompt turns.

## Why the command result type changed

Before this PR, command handling returned a boolean:

```ts
true  // handled as a command
false // not a command
```

That was enough for local commands like `/status`, `/skills`, `/mcp`, and `/logout`.

`/review` is different: it is a slash command, but it also starts a real Codex turn through `review/start`. After that turn completes, the ACP server needs to know whether it was interrupted so it can return the same `cancelled` response used by normal prompt turns.

So command handling now returns a small typed result:

```ts
type CommandHandlingResult =
  | { handled: false }
  | { handled: true; turnCompleted?: TurnCompletedNotification };
```

Existing commands still return `{ handled: true }`, preserving their behavior. `/review` returns `{ handled: true, turnCompleted }`, allowing `CodexAcpServer.prompt()` to propagate interruption and event-handler failures correctly.

This keeps the turn-specific behavior explicit without special-casing `/review` outside the command layer.

## Out of scope

- `/review <custom instructions>`
- branch or commit review selection
- detached review threads
- ACP-native review UI
- regenerating or modifying `src/app-server` protocol types

## Verified

- `npm run typecheck`
- `npm test` (176 passed, 27 skipped)
- `npm run build`
- Manual IntelliJ ACP smoke test:
  - verified `/review` appears and runs against uncommitted changes
  - verified review output is rendered in chat

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -28,6 +28,7 @@ import type {
     GetAccountResponse,
     ListMcpServerStatusResponse,
     Model,
+    ReviewTarget,
     SkillsListParams,
     SkillsListResponse,
     Thread,
@@ -472,6 +473,14 @@ export class CodexAcpClient {
 
     markTurnStale(params: { threadId: string, turnId: string }): void {
         this.codexClient.markTurnStale(params.threadId, params.turnId);
+    }
+
+    async runReview(threadId: string, target: ReviewTarget): Promise<TurnCompletedNotification> {
+        return await this.codexClient.runReview({
+            threadId,
+            target,
+            delivery: "inline",
+        });
     }
 
     async listSkills(params?: SkillsListParams): Promise<SkillsListResponse> {
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -475,12 +475,16 @@ export class CodexAcpClient {
         this.codexClient.markTurnStale(params.threadId, params.turnId);
     }
 
-    async runReview(threadId: string, target: ReviewTarget): Promise<TurnCompletedNotification> {
+    async runReview(
+        threadId: string,
+        target: ReviewTarget,
+        onTurnStarted?: (turnId: string) => void
+    ): Promise<TurnCompletedNotification> {
         return await this.codexClient.runReview({
             threadId,
             target,
             delivery: "inline",
-        });
+        }, onTurnStarted);
     }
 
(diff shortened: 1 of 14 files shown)


anna239:
@gabrielbergamini please rebase your branch on top of the fresh main

gabrielbergamini:
Rebased onto the current `main` and added a focused lifecycle hardening follow-up in `a2752e0`.

After comparing the overlapping review implementation in #181, I adopted one relevant improvement while keeping this PR's narrower scope: the `review/start` turn ID is now stored as `sessionState.currentTurnId`, so ACP cancel/session close can interrupt an in-progress `/review`.

I added a regression test that starts a pending review, issues ACP cancellation, and verifies `turn/interrupt` targets the review turn.

Fresh verification:

- `npm run typecheck`
- `npm test` (176 passed, 27 skipped)
- `npm run build`

benbrandt:
I believe this is now handled in #181

PR #154, opened 2026-05-21
feat: expose model config options

SteffenDE:
Add session config options for mode, model, and reasoning effort, and route session/set_config_option changes through the same session state used for prompts.

This mirrors what the old codex-acp adapter does. The "legacy" models are a combination of model + effort, but the config options expose those separately.

Assisted-by: GPT-5.5


benbrandt:
We definitely need this and should do this. Especially since models are planned to be deprecated / removed in v2.

AlexandrSuhinin:
@benbrandt Agree, just want to double-check that this won't affect IDEA integration.

AlexandrSuhinin (approved):

PR #156, opened 2026-05-22
Fix broken skills (additionalDirectories) in 0.130.0 

NikolaiSviridov:


AlexandrSuhinin:
Why this is done on top of update to 0.130.0? I would expect that we first change skills to new implementation and the do codex update (probably to latest 0.132.0) and check that all tests are still green.

src/CodexAcpClient.ts
```diff
@@ -206,7 +208,14 @@ export class CodexAcpClient {
     }
 
     async resumeSession(request: acp.ResumeSessionRequest): Promise<SessionMetadata> {
+        const additionalRoots = readAdditionalRoots(request._meta, request.additionalDirectories);
```
AlexandrSuhinin: Please extract duplication.

src/CodexAcpClient.ts
```diff
@@ -684,6 +719,26 @@ interface GatewayConfig {
     }
 }
 
+function readAdditionalRoots(
+    meta: Record<string, unknown> | null | undefined,
```
AlexandrSuhinin, line 720: What is this meta? Can we create a type for better readability?
| undefined -> meta?: Type

NikolaiSviridov:
@AlexandrSuhinin skills were broken in 0.130.0 and there is a new logic related to skill handling in it (recomended by openai). So, I wanted to merge update 0.130.0 and update to 0.132.0 after.

Can move it to 0.132.0

src/CodexAcpClient.ts
```diff
@@ -206,7 +208,14 @@ export class CodexAcpClient {
     }
 
     async resumeSession(request: acp.ResumeSessionRequest): Promise<SessionMetadata> {
+        const additionalRoots = readAdditionalRoots(request._meta, request.additionalDirectories);
         await this.refreshSkills(request.cwd);
+        await installAdditionalRootSkillMarketplaces({
```
AlexandrSuhinin, line 228: Why this function is not in PR?

src/SkillDirectoryParser.ts
```diff
@@ -0,0 +1,266 @@
...
+    directory: string;
+    depth: number;
+};
+
+/**
+ * Reads Codex skills from a local skills directory using the same core filesystem conventions as
+ * codex app-server: skills are declared by files named exactly `SKILL.md`, with metadata in YAML
+ * frontmatter and optional UI metadata in `agents/openai.yaml`.
+ */
+export async function parseSkillsFromDirectory(rootDir: string): Promise<ParsedSkillLoadResult> {
```
AlexandrSuhinin, line 49: Where this function is used?

src/CodexAppServerClient.ts
```diff
@@ -265,6 +273,22 @@ export class CodexAppServerClient {
         return await this.sendRequest({ method: "skills/list", params });
     }
 
+    async marketplaceAdd(params: MarketplaceAddParams): Promise<MarketplaceAddResponse> {
```
AlexandrSuhinin, line 277: Where these functions are used?

src/CodexAcpClient.ts
```diff
@@ -52,6 +53,7 @@ export class CodexAcpClient {
     private readonly config: JsonObject;
     private readonly modelProvider: string | null;
     private gatewayConfig: GatewayConfig | null;
+    private readonly sessionAdditionalRoots = new Map<string, string[]>();
```
AlexandrSuhinin: what are these strings? Such code is hard to undersand and maintain.

AlexandrSuhinin (changes requested):

AlexandrSuhinin:
Please add e2e tests if they are missing.

AlexandrSuhinin:
So there is no marketplace API in 0.128.0 version and you're dependent on update to 0.130.0+?

AlexandrSuhinin:
Also I don't really understand if it's a bug, but this review doesn't include all related changes.

src/CodexAppServerClient.ts
NikolaiSviridov: in `installGeneratedSkillPlugin`

src/SkillDirectoryParser.ts
NikolaiSviridov: in `createMarketplaceFromSkills`

src/CodexAcpClient.ts
NikolaiSviridov: it's a reverted code which was deleted by agent in 0.130.0 update. Original author @slapoguzov .

Extracts "additionalRoots" from request._meta

src/CodexAcpClient.ts
NikolaiSviridov: `LocalSkillMarketplace.ts`

src/CodexAcpClient.ts
NikolaiSviridov: sessionId to [additionalRoots]

src/CodexAcpClient.ts
NikolaiSviridov: It has been here for a long time. I didn't remove it because you asked me not to mix new features/fixes with refactoring.

I will extract it

src/CodexAppServerClient.ts
AlexandrSuhinin: Switched to IDE review, GH is inconvinient.

src/CodexAcpClient.ts
AlexandrSuhinin: You can remain it as is. We can extract duplicates from the whole methods later. 
To clarify I meant to avoid unrelated refactorings that add unrelated diffs to the commit (e.g. rename, move, etc) that are optional.
The goal is to keep PR focused and easy to understand.

src/__tests__/CodexACPAgent/e2e/acp-e2e.test.ts
```diff
@@ -103,4 +106,53 @@ describeE2E("E2E tests", () => {
...
+        expect(await fixture.hasMarketplace(HELLO_WORLD_MARKETPLACE_NAME)).toBe(true);
+
+        await fixture.removeMarketplace(HELLO_WORLD_MARKETPLACE_NAME);
+
+        expect(await fixture.hasMarketplace(HELLO_WORLD_MARKETPLACE_NAME)).toBe(false);
+    });
 });
+
+async function prepareHelloWorldAdditionalRoot(fixture: SpawnedAgentFixture): Promise<string> {
+    const additionalRoot = fixture.copyAdditionalRootFixture(HELLO_WORLD_ADDITIONAL_ROOT_FIXTURE);
```
ishulgin: Seems like there is no need to store SKILL.md in repo and introduce copying of it. You can extend `SpawnedAgentFixture#writeSkill` to support a custom dir.

src/CodexAcpServer.ts
```diff
@@ -123,6 +123,19 @@ export class CodexAcpServer implements acp.Agent {
                 await this.unstable_logout({});
                 return {};
             }
+            case "marketplace/list": {
```
AlexandrSuhinin: Was it already discussed and approved to use these extension methods instead of previous _meta approach?
Are we ready that old IDE plugin version won't support this?

AlexandrSuhinin (changes requested):

AlexandrSuhinin:
Please rebase this branch on master if marketplace events are supported in 0.128.0.
Reviewing and merging this will be too error-prone otherwise.

src/__tests__/CodexACPAgent/e2e/acp-e2e.test.ts
NikolaiSviridov: removed `copyAdditionalRootFixture`

src/CodexAcpServer.ts
NikolaiSviridov: Please clarify what you mean. There was no _meta field for such a case; it's used in the e2e test to track whether the marketplace is removed or added. It should not break anything on the old IDEA plugin side

AlexandrSuhinin:
I tested this fix, and it actually installs this plugin into CLI, plugin installation isn't virtual, plugin and skills are visible in the CLI.
I believe that this solution doesn't work for us in any case. We are not ready to add plugins in codex TUI. (And if we are ready, we can do it much easier).

I suggest to close this PR.

PR #164, opened 2026-05-28
fix: read model provider from config on session resume

xiatg:
## Summary
- Reverts PR #139 which introduced `getResumeModelProvider()` with a hardcoded fallback to `"openai"`
- The previous change forced the model provider to default to `"openai"` when none was explicitly passed, overriding the user's configured provider in `~/.codex/config.toml`
- Per the [official configuration guide](https://developers.openai.com/codex/config-advanced#custom-model-providers), users can set their model provider in `~/.codex/config.toml` — the hardcoded fallback bypasses this

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -205,7 +205,7 @@ export class CodexAcpClient {
         const response = await this.codexClient.threadResume({
             config: await this.createSessionConfig(request.cwd, request.mcpServers ?? []),
             cwd: request.cwd,
-            modelProvider: this.getResumeModelProvider(),
+            modelProvider: await this.getResumeModelProvider(),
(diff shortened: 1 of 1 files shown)


AlexandrSuhinin (changes requested):
Proper fix should set provider from the config then. Reverting this PR will break resume on sessions if provider was changed.

xiatg:
Updated to use `getCurrentModelProvider()` in `getResumeModelProvider()`, which reads `model_provider` from the config as a fallback instead of only checking the environment variable. The `"openai"` fallback is kept for the case where no provider is configured anywhere.

benbrandt:
I think #224 is closer to what we are looking for, but thank you!

PR #167, opened 2026-06-01
Suppress diff exceptions

AlexandrSuhinin:


src/CodexToolCallMapper.ts
```diff
@@ -257,56 +258,84 @@ function createSearchTitle(query: string | null, path: string | null): string {
...
-
-        return {
-            type: "diff",
-            oldText: oldContent,
-            newText: "",
-            path: change.path,
-            _meta: {
-                kind: "delete",
-            }
+    try {
```
ishulgin, line 261: Looks like unified diff handling for add/delete are redundant - see `fn format_file_change_diff(change: &FileChange) -> String {` https://github.com/openai/codex/blob/67b805fc111706aca5b32d465c94d95659bab6aa/codex-rs/app-server-protocol/src/protocol/item_builders.rs#L302

src/CodexToolCallMapper.ts
```diff
@@ -257,56 +258,84 @@ function createSearchTitle(query: string | null, path: string | null): string {
...
+        }
+    }
+}
+
+/**
+ * Fix unified diff content corrupted by codex agent.
+ * Removes synthetic "Moved to" from the end.
+ */
+function recoverCorruptedDiff(diff: string): string {
+    return diff.replace(/\nMoved to: .*$/, "");
```
ishulgin: It actually adds 2 \n

src/CodexToolCallMapper.ts
```diff
@@ -257,56 +258,84 @@ function createSearchTitle(query: string | null, path: string | null): string {
...
+
+    const unifiedDiff = recoverCorruptedDiff(change.diff);
+    const newContent = applyPatch(oldContent, unifiedDiff);
+    if (!newContent) return null;
+
+    return {
+        type: "diff",
+        oldText: oldContent,
         newText: newContent,
         path: change.path,
```
ishulgin: What if file was actually moved? Don't we want to use PatchChangeKind#moved_path?

AlexandrSuhinin: Actually I think that it's impossible to express move in a single diff event. We probably need to split it into delete + add pair.

src/CodexToolCallMapper.ts
AlexandrSuhinin: I'm still not sure that it never leaks via content. We have explicit commits and test for add/delete recovery. I tried to avoid any semantic change here and suggest putting this cleanup out of the scope of this PR.

src/CodexToolCallMapper.ts
AlexandrSuhinin: image

Confirmed
image

src/CodexToolCallMapper.ts
```diff
@@ -257,56 +258,84 @@ function createSearchTitle(query: string | null, path: string | null): string {
...
+        newText: newText,
+        path: change.path,
+        _meta: {
+            kind: "add",
+        },
+    };
+}
+
+async function createUpdateFileContent(change: FileUpdateChange): Promise<ToolCallContent | null> {
+    const oldContent = await readFile(change.path, { encoding: "utf8" }).catch(() => null);
```
SteffenDE: I think I'm seeing an error here where, when running the agent on bypass permissions, the edit is already applied to the file when reading the file here, so the applyPatch fails and the "Editing files" ACP tool call is sent with empty content.

src/CodexToolCallMapper.ts
AlexandrSuhinin: We need to discuss how to properly send moved files. For now we send moved_path + oldText/newText.
Let's fixup this in following PR.

src/CodexToolCallMapper.ts
```diff
@@ -257,93 +258,98 @@ function createSearchTitle(query: string | null, path: string | null): string {
...
-            kind: change.kind.type,
+            kind: "add",
         },
     };
 }
 
-function isUnifiedDiff(content: string): boolean {
-    return content.startsWith("--- ") || content.includes("\n--- ");
+async function createUpdateFileContent(change: FileUpdateChange): Promise<ToolCallContent | null> {
+    if (change.kind.type !== "update") return null;
```
ishulgin, line 289: We don't have similar guards in other execution branches. Let's drop it?

ishulgin (approved):

PR #171, opened 2026-06-03
Do not send reason as content on permission request

SteffenDE:
According to the specification, the permissionRequest contains a `ToolCallUpdate`, which - again according to the spec - "Replace[s] the content collection."

https://agentclientprotocol.com/protocol/v1/schema#param-content-13

This means that if the agent sends a tool call with a diff for file edits and then the permission request, a typical client would replace the the diff with just the reason text.

I'd say that either ACP should add a new field `reason` to `RequestPermissionRequest`, or we use the existing `_meta` field to include the reason for clients that want to handle it specially.

(cc @benbrandt)

This PR moves the reason into _meta.

This also solves the problem I had initially in https://github.com/agentclientprotocol/codex-acp/pull/126, where I didn't see any file content and Codex wrongly assumed that it needs to handle more events.

### DIFF
--- a/src/CodexApprovalHandler.ts
+++ b/src/CodexApprovalHandler.ts
@@ -7,7 +7,6 @@ import type {
     FileChangeRequestApprovalParams,
     FileChangeRequestApprovalResponse
 } from "./app-server/v2";
-import type {ToolCallContent} from "@agentclientprotocol/sdk/dist/schema/types.gen";
 import {logger} from "./Logger";
 import {stripShellPrefix} from "./CodexEventHandler";
 import {ApprovalOptionId} from "./ApprovalOptionId";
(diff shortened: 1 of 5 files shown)


src/CodexApprovalHandler.ts
```diff
@@ -62,47 +61,32 @@ export class CodexApprovalHandler implements ApprovalHandler {
...
             sessionId,
             toolCall: {
                 toolCallId: params.itemId,
                 kind: "execute",
                 status: "pending",
-                content: reasonContent ? [reasonContent] : null,
                 rawInput: params.command ? { command: stripShellPrefix(params.command), cwd: params.cwd } : null,
             },
             options: APPROVAL_OPTIONS,
+            _meta: { codex: { params } }
```
SteffenDE, line 73: Of course, we could also cherry-pick specific fields only.

benbrandt:
Maybe silly question... but content is an array, can we not have both?

benbrandt (approved):
Either way, I think _not overwriting_ the diff is the right choice here

SteffenDE:
I'm not sure what you're trying to say? If the permissionRequest contains `content` in the tool call update, the value from the permission request should replace the existing content (e.g. diffs). What codex-acp could do is send the previous content and append the reason as text as another item in the content, but then the question is if this should remain in the content forever or a future update (e.g. completed) should remove it again, but tracking that seems unnecessarily complex.

I'd argue that the reason for a tool call request is something different than the tool call's content, so it shouldn't be mixed.

PR #177, opened 2026-06-05
feat: Implement session/close

benbrandt:
- **feat: Add session close support**
- **Resolve interrupted turns after failed interrupts**
- **Suppress MCP startup updates during session close**
- **Handle session close races cleanly**
- **Close stale session opens on generation mismatch**
- **Prevent stale session cleanup races**
- **Block session reopen during stale cleanup**
- **Fence stale session and turn cleanup on close**

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -258,6 +258,14 @@ export class CodexAcpClient {
         };
     }
 
+    async closeSession(sessionId: string): Promise<void> {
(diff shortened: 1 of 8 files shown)


benbrandt:
@AlexandrSuhinin trying a simpler version again. Would be great to get in, this + config options are the main blockers right now for Zed adoption

AlexandrSuhinin (approved):

PR #178, opened 2026-06-05
Fix stale file read after patch causing empty diff content

SteffenDE:
Also ensure events are always handled in order.

https://github.com/agentclientprotocol/codex-acp/pull/167 fixed some diff errors, but there was still a race condition when using codex in full access mode, where the file was already updated when codex-acp read it, making the patch fail to apply and returning empty content:

https://github.com/agentclientprotocol/codex-acp/blob/e698674ae835d80fdfe5dbffa1eaae9db549ba11/src/CodexToolCallMapper.ts#L294-L297

Additionally, because each event was processed individually, the `await readFile` could cause the `item/completed` event to be processed and sent as `tool_call_update` via ACP before the initial `tool_call` was sent.

### DIFF
--- a/src/CodexToolCallMapper.ts
+++ b/src/CodexToolCallMapper.ts
@@ -294,7 +294,18 @@ async function createUpdateFileContent(change: FileUpdateChange): Promise<ToolCa
     const oldContent = await readFileContent(change.path);
     if (oldContent !== null) {
         const patchedContent = applyPatch(oldContent, unifiedDiff);
-        if (patchedContent === false) return null;
+        if (patchedContent === false) {
+            // If Codex runs in full access mode, the file might already be patched.
+            // we can verify this by checking if the reverted patch applies.
+            const revertedPatch = revertPatch(unifiedDiff);
+            if (revertedPatch) {
+                const revertedContent = applyPatch(oldContent, revertedPatch);
+                if (revertedContent !== false) {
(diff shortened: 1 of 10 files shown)


src/CodexToolCallMapper.ts
```diff
@@ -294,7 +294,18 @@ async function createUpdateFileContent(change: FileUpdateChange): Promise<ToolCa
...
+            // we can verify this by checking if the reverted patch applies.
+            const revertedPatch = revertPatch(unifiedDiff);
+            if (revertedPatch) {
+                const revertedContent = applyPatch(oldContent, revertedPatch);
+                if (revertedContent !== false) {
+                    return createUpdateDiffContent(change.path, revertedContent, oldContent);
+                }
+            }
+            return null;
+        }
```
SteffenDE, line 308: This one handles the empty content. The other changes are all for ensuring correct message order.

src/__tests__/CodexACPAgent/data/mcp-tool-completed-with-logs.json
```diff
@@ -4,12 +4,23 @@
     {
       "sessionId": "test-session-id",
       "update": {
-        "sessionUpdate": "tool_call_update",
+        "sessionUpdate": "tool_call",
```
SteffenDE, line 7: Those test fixtures actually encoded the bad order.

AlexandrSuhinin (approved):

PR #182, opened 2026-06-08
feat: Stream reasoning events as agent thoughts

benbrandt:
We weren't showing any reasoning/thinking before, but this also allows
for streaming in the reasoning deltas if the backend supports it.

### DIFF
--- a/src/CodexEventHandler.ts
+++ b/src/CodexEventHandler.ts
@@ -20,6 +20,9 @@ import type {
     ItemStartedNotification,
     ThreadItem,
     ModelReroutedNotification,
+    ReasoningSummaryPartAddedNotification,
+    ReasoningSummaryTextDeltaNotification,
+    ReasoningTextDeltaNotification,
     ThreadGoalClearedNotification,
     ThreadGoalUpdatedNotification,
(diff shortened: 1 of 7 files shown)


src/CodexEventHandler.ts
```diff
@@ -305,11 +323,46 @@ export class CodexEventHandler {
...
+        return {
+            sessionUpdate: "agent_message_chunk",
+            content: {
+                type: "text",
+                text,
+            }
+        };
+    }
+
+    private createContextCompactedEvent(): UpdateSessionEvent {
```
AlexandrSuhinin, line 426: Let's remove changes related to PR with /review & /compact support

AlexandrSuhinin (changes requested):

AlexandrSuhinin (approved):

AlexandrSuhinin (changes requested):
Let's keep PRs focused on one change to simplify reviewing.

benbrandt:
Cleaned up. apologies for the mix up, must have been moving between branches too quickly

AlexandrSuhinin (approved):

PR #184, opened 2026-06-09
fix: Handle more web search events

benbrandt:


src/CodexToolCallMapper.ts
```diff
@@ -190,6 +190,64 @@ export function createFuzzyFileSearchComplete(
...
+            const pattern = action.pattern ? ` for '${action.pattern}'` : "";
+            const url = action.url ? ` in ${action.url}` : "";
+            return `Find in page${pattern}${url}`.trim();
+        }
+        case "other":
+            return "Web search";
+    }
+}
+
+function createWebSearchRawInput(item: ThreadItem & { type: "webSearch" }): {
```
AlexandrSuhinin: Maybe it's worth to introduce type for return value. Or completely inline function.

AlexandrSuhinin (approved):

PR #188, opened 2026-06-09
Handle file change patch updates

benbrandt:


SteffenDE:
I may be misremembering, but I think Codex flagged my usage of `patchUpdated` quoting the docs

> item/fileChange/patchUpdated - when features.apply_patch_streaming_events is enabled, streams structured file-change snapshots parsed from the model-generated patch before it is executed.

with something about patchUpdated being sent before the item, so it could happen that we create a tool_call_update before the initial tool_call.

I tried to see if I can get Codex to actually emit a patchUpdated, but I did not see it to verify...

PR #191, opened 2026-06-09
feat: Support ACP additional directories

benbrandt:


src/CodexAcpClient.ts
```diff
@@ -328,17 +342,25 @@ export class CodexAcpClient {
         return this.getModelProvider() ?? "openai";
     }
 
-    private async refreshSkills(cwd: string, meta?: Record<string, unknown> | null): Promise<void> {
+    private async refreshSkills(
+        cwd: string,
+        additionalDirectories: string[],
```
AlexandrSuhinin: Maybe we can unifiy additional directories with meta?
They have the same semantics

AlexandrSuhinin:
Essentially we only populate write permissions to these additional roots? Do we expect that agent will know that it can/should work in these directories from the prompt?
I'm not sure what problem we solve here.

AlexandrSuhinin (changes requested):

benbrandt:
The goal is to support the same behavior as `--add-dir` in the codex CLI which allows you to have a session with multiple directories. We allow for multi-directory projects in zed, so this allows us to pass them all in to the session

AlexandrSuhinin (approved):

PR #196, opened 2026-06-12
Add /compact command for context compaction

SteffenDE:
Assisted-by: GPT-5.5

Reports the compaction as a running tool call. Could also be a regular assistant message, but I think the tool call works quite well.


AlexandrSuhinin:
Fixed in #181 ?

SteffenDE:
Yes, I missed the other PR, sorry about that!

PR #199, opened 2026-06-14
Keep configured model id for custom providers instead of forcing the default

scottzx:
## Problem

When `~/.codex/config.toml` selects a custom provider with a model the built-in catalog doesn't know about (e.g. a self-hosted or third-party endpoint), every turn is sent with the **wrong** model id and the request fails:

```json
{"error":{"message":"unknown model 'gpt-5.5'","code":"invalid_prompt"}}
```

…even though the same config works fine through the Codex CLI.

## Root cause

`CodexAcpClient.createModelId()` resolves the configured/returned model id against `listModels()` and, when it's missing, substitutes `availableModels.find(m => m.isDefault)`:

```ts
const selectedModel =
    availableModels.find(m => m.id === modelId) ??
    availableModels.find(m => m.isDefault);   // ← discards the configured model
```

Custom-provider model ids are **not** enumerated by `listModels()`, so the configured model is silently replaced by the built-in default. That default id (`response.model` → `currentModelId`) is then pinned onto every `runTurn({ model })`, so requests to the custom endpoint fail with `unknown model '<default>'`.

The Codex **CLI** handles the identical case differently — it keeps the configured model id and only warns:

> Model metadata for `MiniMax-M3` not found. Defaulting to fallback metadata; this can degrade performance and cause issues.

## Fix

Mirror the CLI: when the model id isn't in the advertised catalog, **keep it** instead of substituting the default. Only fall back to `isDefault` when no model id was supplied at all. Downstream session state already degrades gracefully for unknown models (`supportedReasoningEfforts ?? []`, `inputModalities ?? ["text", "image"]`).

## Testing

- Added two unit tests in `CodexAcpClient.test.ts` for an uncatalogued model id (with an explicit reasoning effort, and defaulting the effort when `null`).
- `vitest run` (createModelId suite) and `tsc --noEmit` pass.
- Verified end-to-end through an ACP client: a session with `model = "MiniMax-M3"` now advertises `MiniMax-M3[high]` (previously `gpt-5.5`) and a prompt turn completes with a real reply from the custom endpoint.


benbrandt (approved):
Thanks!

PR #207, opened 2026-06-16
Allow disabling MCP config filtering

i1bro:


src/CodexAcpClient.ts
```diff
@@ -314,7 +316,9 @@ export class CodexAcpClient {
 
         // Deduplicates new servers against existing config to prevent Codex from deep-merging
         // incompatible field types (e.g., mixing url and stdio schemas).
-        const existingNames = await this.getConfigMcpServerNames(projectPath);
+        const existingNames = this.disableMcpConfigFiltering
```
AlexandrSuhinin: I would create a single function shouldDeduplicateMcpConflicts() that will check for env variable or arg. And would avoid leaking it to constructor, field, etc.

src/CodexAcpClient.ts
AlexandrSuhinin: I would rather check it here, right where we filter mcp's to make the code explicit.

const uniqueServers ...
let mcpSevers ...

if (should deduplicate) {
   filter
} else { 
 ...
}

AlexandrSuhinin (changes requested):
Also we need a test for this.

src/__tests__/CodexACPAgent/CodexAcpClient.test.ts
```diff
@@ -328,6 +328,58 @@ describe('ACP server test', { timeout: 40_000 }, () => {
         });
     });
 
+    it('passes conflicting ACP MCP servers through when config filtering is disabled', async () => {
```
AlexandrSuhinin: Too many mocks, and we already have test file for such cases: `mcp-config-merge.test.ts`

src/CodexAcpClient.ts
```diff
@@ -716,6 +718,12 @@ function formatUriAsLink(name: string | null | undefined, uri: string): string {
     return uri;
 }
 
+function shouldDeduplicateMcpConflicts(): boolean {
+    const disabledByEnv = process.env["DISABLE_MCP_CONFIG_FILTERING"] === "true";
+    const disabledByArg = process.argv.includes("--disable-mcp-config-filtering");
```
AlexandrSuhinin: Why we add 2 different flags? Let's keep one.

AlexandrSuhinin (changes requested):

AlexandrSuhinin (approved):

PR #208, opened 2026-06-16
Load full thread history on session resume

### LINKED ISSUE #206: session/load restores text history but misses historical tool calls
When loading an existing session, codex-acp restores prior user/assistant text messages but does not replay historical tool activity.

This makes loaded sessions incomplete compared with the original live session. Clients can see the text conversation, but lose prior commands, MCP/dynamic tool calls, statuses, inputs/outputs, and affected locations.

Observed behavior:

  - session/load emits text history, such as user_message_chunk and agent_message_chunk
  - prior tool activity is not replayed as tool_call / tool_call_update

Expected behavior:

  - session/load should restore the visible conversation history, including prior tool activity
  - after reloading, clients should be able to reconstruct the same conversation view that was visible during the original session

Zed comparison:

The Zed Codex ACP implementation reads Codex rollout history and reconstructs historical tool activity from stored events/items. In contrast, this codex-acp adapter loads a session through the Codex app-server and replays only the thread history returned by that API. In the reproduced case, that app-server thread history contains the text messages but not the prior tool-call records, so codex-acp has no historical tool activity to emit during session/load.

benbrandt:
Closes #206

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -237,6 +237,10 @@ export class CodexAcpClient {
             threadId: request.sessionId,
         });
         onSubscribed?.();
+        const historyResponse = await this.codexClient.threadRead({
+            threadId: response.thread.id,
+            includeTurns: true,
+        });
         const codexModels = await this.fetchAvailableModels();
         const currentModelId = this.createModelId(codexModels, response.model, response.reasoningEffort).toString();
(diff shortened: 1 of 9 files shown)


OldKrab:
Thanks for such fast fix!

AlexandrSuhinin (approved):

PR #212, opened 2026-06-18
fix: include stderr in process exceptions

AlexandrSuhinin:


src/CodexAcpServer.ts
```diff
@@ -142,6 +144,7 @@ export class CodexAcpServer implements acp.Agent {
         this.codexAcpClient = codexAcpClient;
         this.defaultAuthRequest = defaultAuthRequest ?? null;
         this.getExitCode = getExitCode ?? (() => null);
+        this.getRecentStderr = getRecentStderr ?? (() => "");
```
xtmq, line 147: lastError instead of recent maybe?

src/__tests__/CodexACPAgent/process-exit-error.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { once } from 'node:events';
+import * as acp from '@agentclientprotocol/sdk';
+import { startCodexConnection } from '../../CodexJsonRpcConnection';
+import { CodexAppServerClient } from '../../CodexAppServerClient';
+import { CodexAcpClient } from '../../CodexAcpClient';
+import { CodexAcpServer } from '../../CodexAcpServer';
+import { createMockConnections } from './test-utils';
+
+describe('CodexACPAgent - process exit error', () => {
+    it.skipIf(process.platform === 'win32')('includes the crashed process stderr', async () => {
```
xtmq, line 14: maybe add win case also? :)

xtmq (approved):

src/CodexAcpServer.ts
AlexandrSuhinin: It's not actually last, it's all stderr limited to some size.
I'm not sure we can actually get single error - it still can be spitted into a few data events.

src/__tests__/CodexACPAgent/process-exit-error.test.ts
AlexandrSuhinin: We don't have CI on win, also the change is not OS specific. I suggest to keep it as is, otherwise I need to blindly add win case support without actual checking or spend unreasonable time to check it.

PR #224, opened 2026-06-23
fix: use configured model provider on session resume

david-melville:
## Problem

When resuming or loading a session, ACP's `getResumeModelProvider()` returned `"openai"` as a hardcoded fallback whenever `MODEL_PROVIDER` was not set in the environment — even if `~/.codex/config.toml` explicitly set `model_provider = "azure"`.

This is a v1.0.0 regression: Codex itself reads the config correctly (verified via `config/read`), but the ACP adapter overrode it on resume/load.

## Fix

`getResumeModelProvider()` now awaits an async lookup of the configured provider (via `config/read`) before falling back to `"openai"`. Both `resumeSession` and `loadSession` pass the resolved provider.

## Testing

- Added a dedicated test: *uses configured model provider when resuming sessions without an explicit provider*
- All existing tests pass (48 focused, 222 full suite)
- Typecheck and build clean

Related: #164

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -178,7 +178,7 @@ export class CodexAcpClient {
             return sessionModelProvider;
         }
         const settingsModelProvider = await this.codexClient.configRead({includeLayers: false});
-        return settingsModelProvider.config.model_provider ?? null;
+        return settingsModelProvider?.config?.model_provider ?? null;
     }
 
     async logout(): Promise<void> {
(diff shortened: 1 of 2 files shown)


Bingmang:
Thanks for the fix. I've run into the same issue on my side as well. It would be greatly appreciated if this could be merged and released soon. @benbrandt 

Looking forward to seeing this fix included in an upcoming release.

benbrandt (approved):
Thanks!

PR #228, opened 2026-06-23
fix: fall back to inline image data for internal URIs

### LINKED ISSUE #231: Bug: Regression: `image_url` validation fails for local file URIs (`file://` protocol) in v1.0.0
**Describe the bug**
In version 1.0.0, passing a local file URI (e.g., `file:///path/to/file.png`) to the `image_url` field results in a `400 Bad Request` validation error. Previously, local file URIs did not trigger this validation failure.

**To Reproduce**
Send a payload to the ACP containing a local file URI in the `image_url` parameter:

```json
{
        "type": "image",
        "mimeType": "image/png",
        "data": "[redacted image/png payload as base64]",
        "uri": "file:///Users/dmeehan/Desktop/Screenshot.png"
 }
```

**Expected behavior**
The local file URI should either be accepted and resolved correctly, or ignored if inline data is provided. It should not fail schema validation with a `400` error.  It should also handle escaped URI's - e.g. %20 for spaces.

**Actual behavior**
The API returns a `400 Bad Request` response:
```json
{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "code": "invalid_value",
    "message": "Invalid 'input[0].content[0].image_url'. Expected a valid URL, but got a value with an invalid format.",
    "param": "input[0].content[0].image_url"
  },
  "status": 400
}
```

**Environment:**
- **OS**: macOS
- **AgentClientProtocol Codex ACP Version**: 1.0.0

**Additional Context**
- In previous versions of Codex ACP, combining `data` (base64) along with a `file://` URI succeeded. It is possible the `uri` property was previously ignored when `data` was present, and v1.0.0 introduced stricter validation that now rejects the `file` protocol. 
- A temporary workaround is to completely omit the file URI and only send the base64 data property.

kolodziejm:
## Summary

- Fall back to inline ACP image data when an image block carries an internal or otherwise non-fetchable URI.
- Keep existing behavior for `http:`, `https:`, and `data:` image URLs.
- Add a regression test covering `zed:///agent/pasted-image?...` input.

## Motivation

Some ACP clients attach pasted images with inline `data`/`mimeType` plus a client-internal `uri`. For example, Zed can send a pasted image URI such as `zed:///agent/pasted-image?name=Image`. The adapter currently forwards that URI to Codex as an image URL, which later becomes an OpenAI `image_url` value and is rejected because `zed:///...` is not a valid URL for the API.

Using the inline image payload for non-fetchable URI schemes preserves image input while still allowing real external URLs and already-inline `data:` URLs to pass through unchanged.

## Validation

- `npm test -- src/__tests__/CodexACPAgent/CodexAcpClient.test.ts`
- `npm run typecheck`

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -703,7 +703,8 @@ function buildPromptItems(prompt: acp.ContentBlock[]): UserInput[] {
             case "text":
                 return {type: "text", text: block.text, text_elements: []};
             case "image": {
-                const url = block.uri ?? `data:${block.mimeType};base64,${block.data}`;
+                const dataUrl = `data:${block.mimeType};base64,${block.data}`;
+                const url = isSupportedImageUrl(block.uri) ? block.uri : dataUrl;
                 return {type: "image", url};
             }
             case "resource_link":
@@ -731,6 +732,10 @@ function buildPromptItems(prompt: acp.ContentBlock[]): UserInput[] {
 
 function isImageMimeType(mimeType: string | null | undefined): mimeType is string {
     return mimeType?.startsWith("image/") ?? false;
+}
+
+function isSupportedImageUrl(uri: string | null | undefined): uri is string {
+    return typeof uri === "string" && /^(https?:|data:)/i.test(uri);
 }
 
 function formatUriAsLink(name: string | null | undefined, uri: string): string {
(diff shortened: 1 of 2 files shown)


benbrandt (approved):
Thanks!

PR #245, opened 2026-06-25
fix: Track configured auth for turn errors

### LINKED ISSUE #244: AuthRequired error usage not consistent with other implementations
The zed-industries/codex-acp and claude-agent-acp implementations respond with `AuthRequired` (-32000) if there is has been no authentication configured but it is required in order to proceed.

However, in the following cases they respond with `InternalError` (-32603):
- An API key was provided, but it was rejected
- Usage limits were exceeded
- Other authentication/login-related issues occured (I've seen 401 responses result in an `InternalError`, for example)

These `InternalError`s contain `data` fields that provide a message describing why the failure occurred.

This implementation, however, always responds with `AuthRequired` for all of the above cases, and doesn't include any reason for the failure in the error response:
https://github.com/agentclientprotocol/codex-acp/blob/1f58237b08fe4877988352f32d448a380603fe70/src/CodexEventHandler.ts#L593-L595

Could you please consider aligning with the other implementations and only using `AuthRequired` when there is currently no authentication configured? 

Populating the `data` field in the error response to describe the failure would also be very useful.

Many thanks!

benbrandt:
Closes #244

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -217,6 +217,10 @@ export class CodexAcpClient {
         return response.requiresOpenaiAuth && !response.account;
     }
 
+    hasGatewayAuth(): boolean {
+        return this.gatewayConfig !== null;
+    }
+
     async getAccount(): Promise<GetAccountResponse> {
         return this.codexClient.accountRead({refreshToken: false});
     }
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -242,6 +242,7 @@ export class CodexAcpClient {
             sessionId: request.sessionId,
             currentModelId: currentModelId,
             models: codexModels,
+            modelProvider: response.modelProvider,
             currentServiceTier: response.serviceTier as ServiceTier ?? null,
             additionalDirectories,
         }
@@ -268,6 +269,7 @@ export class CodexAcpClient {
             sessionId: request.sessionId,
             currentModelId: currentModelId,
             models: codexModels,
+            modelProvider: response.modelProvider,
             currentServiceTier: response.serviceTier as ServiceTier ?? null,
             thread: historyResponse.thread,
             additionalDirectories,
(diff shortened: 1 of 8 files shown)


anvilpete:
Many thanks for this @benbrandt!

I did some quick testing, and it seems to work well.
However, 401 errors look a bit wonky:

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "error": {
    "code": -32603,
    "message": "Internal error: Reconnecting... 5/5",
    "data": {
      "message": "Reconnecting... 5/5",
      "codexErrorInfo": {
        "responseStreamDisconnected": { "httpStatusCode": 401 }
      },
      "additionalDetails": "unexpected status 401 Unauthorized: Incorrect API key provided: fdsa. You can find your API key at https://platform.openai.com/account/api-keys., url: https://api.openai.com/v1/responses, cf-ray: a11b34cebba09505-LHR, request id: req_79cc489e2bac4c1ea0ecb796ab39ab94"
    }
  }
}
```

In the original Zed implementation, the text now in `additionalDetails` was in `data.message`.

Another difference from the Zed implementation: the top-level `message` previously matched the definition in the JSON-RPC spec exactly, and additional detail was in `data.message`. So the error would have looked like:
```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "error": {
    "code": -32603,
    "message": "Internal error",
    "data": {
      "message": "unexpected status 401 Unauthorized: Incorrect API key provided: fdsa. You can find your API key at https://platform.openai.com/account/api-keys., url: https://api.openai.com/v1/responses, cf-ray: a11b34cebba09505-LHR, request id: req_79cc489e2bac4c1ea0ecb796ab39ab94",
...
```

anvilpete:
(the exact shape probably isn't too important, but it would be nice to avoid the "Reconnecting... 5/5")

benbrandt:
great catch thanks, fixing

PR #268, opened 2026-07-07
Support MCP elicitation if the client does

benbrandt:
Holding off on the ask user question tool until it is less experimental.


OldKrab:
@benbrandt Hi! Could you share why `request_user_input` elicitation support was removed before this PR was merged? I’ve been working on supporting it in a separate PR, but then noticed that you had already explored an implementation here. 

Is something needed from ACP before it can be supported properly? Thanks!

benbrandt:
My understanding is this is only triggered if you enable the global experimental flag, which would have effected more parts of the adapter than I would have liked.
If I was wrong about this we can revisit it

OldKrab:
Thanks for answering! As I understand it, the request_user_input tool is available in plan mode without experimental flags. However, it can be enabled in all modes via the default_mode_request_user_input feature flag, which I often use. I don't understand how this flag could affect other parts of the adapter besides elicitation, but I may be wrong.

benbrandt:
Added back support got handle it if it gets called by codex: https://github.com/agentclientprotocol/codex-acp/pull/291
Did not opt-in to the experimental things though, let me know if it works with your config

OldKrab:
@benbrandt I tried it in my client and it looks great!
image

One minor question: why actual question is sent as `description`? But "Time travel" from screenshot is actual `title` field?

OldKrab:
I added https://github.com/agentclientprotocol/codex-acp/pull/299 with changing structure of elicitation for `request_user_input`, to make it more like codex cli do.

PR #269, opened 2026-07-07
Show skill reads as ACP read tool calls

nikita-ashihmin:
## Summary

Render Codex reads of skill instruction files as skill-specific ACP read tool calls.

## Before

- When the agent loaded/read a skill, Codex emitted a normal `commandExecution` item whose parsed `commandActions` contained a `read` action for a `.../skills/<skill>/SKILL.md` path.
- `codex-acp` mapped every read action the same way, so the UI only saw a generic title like `Read file '/path/to/SKILL.md'`.
- That made skill loading indistinguishable from ordinary file reads, even though the UI wants to show activity such as `Read UI Accessibility skill` separately.

## After

- `createCommandActionEvent()` detects read actions whose path points at a skill instruction file (`*/skills/<skill>/SKILL.md`).
- Those reads are still emitted as standard ACP read tool calls:
  - `sessionUpdate: "tool_call"`
  - `kind: "read"`
  - `title: "Read <skill> skill"`
  - `locations: [{ path: <SKILL.md path> }]`
  - `rawInput: { type: "skill", name: <skill directory>, path }`
- Ordinary file reads are unchanged.

## Why

The agent-side “loaded/read skill” activity is already present in Codex events as a parsed read command for `SKILL.md`; `codex-acp` just treated it like a generic file read. This keeps the ACP protocol shape unchanged while giving Codex-aware UIs a stable way to render skill loading separately from regular file reads (`rawInput.type === "skill"`).

## Tests

- `npx vitest run src/__tests__/CodexACPAgent/command-action-events.test.ts -u`
- `npm run typecheck`
- `npm test`

### DIFF
--- a/src/CodexToolCallMapper.ts
+++ b/src/CodexToolCallMapper.ts
@@ -427,7 +427,23 @@ export function createCommandActionEvent(
 ): AcpToolCallEvent {
     const acpStatus = toAcpStatus(status);
     switch (commandAction.type) {
-        case "read":
+        case "read": {
+            const skillName = skillNameFromSkillPath(commandAction.path);
+            if (skillName) {
+                return {
+                    sessionUpdate: "tool_call",
+                    toolCallId: id,
+                    status: acpStatus,
+                    kind: "read",
+                    title: `Read ${formatSkillName(skillName)} skill`,
+                    locations: [{ path: commandAction.path }],
+                    rawInput: {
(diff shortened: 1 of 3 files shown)


nikita-ashihmin:
Closing this PR because the skill-read display can be handled client-side by interpreting existing read tool calls for SKILL.md paths.

PR #272, opened 2026-07-08
feat: implement configurable LLM providers

Rizzen:
Implement the Configurable LLM Providers RFD


AlexandrSuhinin (approved):
Note: also tested that current acp integration in AI Assistant is not affected.

PR #277, opened 2026-07-09
Capitalize reasoning effort labels

joshkent94:
## Summary
- Capitalize reasoning effort option labels in ACP session config
- Keep reasoning effort values lowercase for protocol/model behavior


benbrandt (approved):
Thanks, third time is the charm :D

PR #279, opened 2026-07-10
Fix duplicate model & effort configuration in ACP session responses

duclvz:
Session lifecycle responses exposed model and reasoning options twice: through the legacy models field and standard ACP configOptions. That ACP non-compliance lead to ACP Client render weird duplicate model and effort options.
image

This fixes removes the legacy field from new, load, and resume responses, making configOptions the single standards-compliant configuration surface. Tests and E2E fixtures were updated accordingly.

### DIFF
--- a/src/AcpExtensions.ts
+++ b/src/AcpExtensions.ts
@@ -1,23 +1,9 @@
 import type {
     ClientContext,
-    LoadSessionResponse,
-    NewSessionResponse,
-    ResumeSessionResponse,
     SessionId,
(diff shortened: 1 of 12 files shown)


benbrandt:
Which ACP client is this? I think they should not render models if config options are present.

JetBrains still needs the models so we will leave it.

duclvz:
Hi, this ACP Client is my own project, so the legacy models are intentional. It’s the ACP Client’s responsibility to render them correctly when config options are present. Thanks for clarifying!

PR #281, opened 2026-07-10
Forward unrecognized slash commands to Codex

duclvz:
Pass slash commands not handled by the ACP adapter through as raw prompts, allowing Codex to resolve skills freely. Built-in commands continue to be handled locally.

### DIFF
--- a/src/CodexCommands.ts
+++ b/src/CodexCommands.ts
@@ -254,8 +254,8 @@ export class CodexCommands {
                 return { handled: true };
             }
             default:
-                await this.sendUnknownCommandMessage(commandName, sessionId);
-                return { handled: true };
+                // Let Codex resolve unrecognized commands as raw prompts.
+                return { handled: false };
         }
     }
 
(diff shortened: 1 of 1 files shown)


benbrandt (commented):
@AlexandrSuhinin Maybe you all had this this way for a reason? But I think this change seems reasonable

AlexandrSuhinin:
> @AlexandrSuhinin Maybe you all had this this way for a reason? But I think this change seems reasonable

We initially supported skills with the $ prefix: the same way they are supported in CLI.
Plus I believe that we want to have a completion for commands & skills in the clients. So acp implementation have to announce available options for this.
Also this is a way to advertise available commands.

But anyway I'm not against this change, probably it is still better than throwing an error.

benbrandt (approved):

benbrandt:
Yeah I am also not sure why this is needed to resolve skills (in the PR description that is the motivation...)
but I think this change is ok...

PR #293, opened 2026-07-13
Add plan and goal command actions

nikita-ashihmin:
## Summary
- expose `collaboration_mode` as an ACP session config option and restore it when sessions are resumed or loaded
- advertise `/plan` through neutral `_meta.commandAction` metadata as a stateful `setConfigOption` action
- handle `/plan` locally as a toggle without starting a model turn and publish `config_option_update`
- apply Codex collaboration mode through `thread/settings/update`
- advertise `/goal` as a stateful `prefixPrompt` action so clients collect an objective and send `/goal <objective>`
- publish authoritative goal snapshots in session metadata, including stable identity, status, elapsed time, and an advertised out-of-band control method
- restore or clear the current goal when an existing session is resumed or loaded, while preventing stale reads from replacing newer live updates
- support typed out-of-band Pause and Clear controls while the goal turn is active

## Why

ACP available commands describe command names and arguments, but that is not enough for clients to distinguish a raw slash command from a client-side interaction. Plan changes persistent session configuration; Goal changes how the next prompt is composed, starts a goal turn, and then becomes durable session state.

The optional `_meta.commandAction` hint lets capable clients provide those interactions without hard-coding Codex command names. Goal state is published through session metadata so clients can render and restore a persistent goal panel. The advertised control method avoids trying to start a second ACP prompt while the goal turn is active. Start and Resume continue through the Goal prompt path because they create a model turn; Pause and Clear use the control extension because they do not.

Clients that do not understand these optional metadata fields continue to treat both entries as ordinary slash commands.

## Tests
- `npm test` (310 passed, 28 skipped)
- `npm run typecheck`
- `npm run build`


Ololoshechkin (approved):

src/AcpExtensions.ts
```diff
@@ -7,6 +7,7 @@ import type {
 } from "@agentclientprotocol/sdk";
 
 export const LEGACY_SET_SESSION_MODEL_METHOD = "session/set_model";
+export const GOAL_CONTROL_METHOD = "session/goal_control";
```
Ololoshechkin: @nikita-ashihmin it's non-standard method in ACP, isn't it?

Ololoshechkin: Couldn't find it in ACP spec, so it looks like an extension to me.  The protocol reserves any method name starting with an underscore (_) for custom extensions. See https://agentclientprotocol.com/protocol/v1/extensibility

src/CodexAcpServer.ts
```diff
@@ -853,6 +901,67 @@ export class CodexAcpServer {
...
+    private async publishCurrentGoal(
+        sessionState: SessionState,
+        sessionGeneration: number,
+        force: boolean,
+    ): Promise<void> {
+        const requestRevision = ++sessionState.goalRevision;
+        const goal = await this.runWithProcessCheck(() => this.codexAcpClient.getGoal(sessionState.sessionId));
+        const snapshot = goal === null ? null : toThreadGoalSnapshot(goal);
+        if (!this.goalPublishIsCurrent(sessionState, sessionGeneration)
+            || sessionState.goalRevision !== requestRevision) {
```
Ololoshechkin, line 929: Is there a test for this thread matching by `goalRevision`?

Ololoshechkin (commented):
@nikita-ashihmin please check the `session/goal_control` method -- if it's an extension, please use https://agentclientprotocol.com/protocol/v1/extensibility guidelines

src/AcpExtensions.ts
nikita-ashihmin: Fixed in 41ceacf: the custom request is now namespaced as _codex/session/goal_control, so it starts with the underscore reserved for ACP extension methods. The advertised controlMethod metadata and snapshots were updated as well.

src/CodexAcpServer.ts
nikita-ashihmin: Yes — the test "ignores an older goal refresh that completes after a newer refresh" exercises this exact goalRevision race by resolving the newer refresh first and the stale one afterward, then asserting that only the current goal is published. I reran it as part of the full suite (310 tests passed).

Ololoshechkin (approved):

agentcooper:
@benbrandt @Ololoshechkin Shall we standardize `collaboration_mode` session config option category?

Current spec [says](https://agentclientprotocol.com/protocol/v1/session-config-options#option-categories):
> Category names that do not begin with _ are reserved for the ACP spec.

PR #309, opened 2026-07-16
Steering implementation

Rizzen:


src/CodexAcpServer.ts
```diff
@@ -861,6 +873,145 @@ export class CodexAcpServer {
         };
     }
 
+    async steerSessionWithFallback(params: SessionSteerRequest): Promise<SessionSteeringResponse> {
+        const previousRequest = this.steeringRequests.get(params.sessionId) ?? Promise.resolve();
+        let releaseRequest: () => void = () => {};
+        const requestCompleted = new Promise<void>((resolve) => {
```
Ololoshechkin: This queue part is a bit too complex. Let's refactor to a normal queue and add some comments

Ololoshechkin: Please consider one of the following options:
a) No `steeringRequests` at all -- I'm not sure if we need to wait. Please check
b) Just keep atomic reference to one single "lastRequest" and wait for it

src/CodexAcpServer.ts
```diff
@@ -861,6 +873,145 @@ export class CodexAcpServer {
...
+
+    private async performSteeringRequest(params: SessionSteerRequest): Promise<SessionSteeringResponse> {
+        logger.log("Steering session requested", {
+            sessionId: params.sessionId,
+            prompt: params.prompt,
+        });
+        const sessionState = this.getSessionState(params.sessionId);
+
+        if (!sessionState.supportedInputModalities.includes("image") && params.prompt.some(b => b.type === "image")) {
+            throw RequestError.invalidRequest("The current model does not support image input");
```
Ololoshechkin: Is there any catch for this throw on the top level of the wrapper? What will the user get?
It is important to make sure that if user sends some incorrect steering request mid-execution, codex still proceeds it's session and IDE is still connected.
Ideally -- let's think about testing

src/CodexAcpServer.ts
```diff
@@ -861,6 +873,145 @@ export class CodexAcpServer {
...
+        });
+        return {outcome: "injected"};
+    }
+
+    private async startNewTurnFromSteering(
+        params: SessionSteerRequest,
+        previousPrompt: ActivePrompt | undefined,
+    ): Promise<SessionSteeringResponse> {
+        await previousPrompt?.completion;
+        if (this.sessionIsClosing(params.sessionId)) {
```
Ololoshechkin, line 1018: This condition is always true -- plese fix. Otherwise, `throw RequestError.invalidRequest` will happen always

src/CodexAcpServer.ts
```diff
@@ -861,6 +873,145 @@ export class CodexAcpServer {
...
+            throw RequestError.invalidRequest(`Session ${params.sessionId} is closing`);
+        }
+
+        let resolveTurnStarted: () => void = () => {};
+        const turnStarted = new Promise<void>((resolve) => {
```
Ololoshechkin: Let's simplify this part please

src/CodexAcpServer.ts
```diff
@@ -861,6 +873,145 @@ export class CodexAcpServer {
...
+                threadId: params.sessionId,
+                turnId,
+                prompt: params.prompt,
+            }));
+        } catch (err) {
+            await this.codexAcpClient.waitForSessionNotifications(params.sessionId);
+            if (sessionState.currentTurnId === turnId && !this.isNoActiveTurnToSteerError(err)) {
+                throw err;
+            }
+            return await this.startNewTurnFromSteering(params, activePrompt);
```
Ololoshechkin: Would love to see some test for this behaviour (but we can add tests later in a separate PR)

Ololoshechkin (changes requested):

Ololoshechkin (approved):

PR #311, opened 2026-07-17
fix(tool-call-edit-diff): emit compact file-edit diffs and raw io

duclvz:
## Problem

File-edit tool calls expand small changes into whole-file diffs and omit useful ACP metadata. This creates noisy output and can drop edits when the source file is unavailable or already patched.

image

## Solution

- Emit compact, per-hunk diffs without reading files from disk
- Add `locations`, `rawInput`, and `rawOutput`
- Map `item/fileChange/patchUpdated` events
- Keep session-history replay consistent

## Test plan

- [x] `npm run typecheck`
- [x] `npm test` (312 passing)
- [x] Regression coverage for invalid patches, unavailable files, patch updates, and event ordering

### DIFF
--- a/src/CodexEventHandler.ts
+++ b/src/CodexEventHandler.ts
@@ -39,6 +39,8 @@ import {
     createContextCompactionCompleteUpdate,
     createContextCompactionStartUpdate,
     createDynamicToolCallUpdate,
+    createFileChangeCompleteUpdate,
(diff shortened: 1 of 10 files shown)


duclvz:
Could someone please take a look when you have a chance? Your feedback would be greatly appreciated to help move this change forward. Thanks!

PR #339, opened 2026-07-28
Support ACP audio prompts

shreyas-shinde:
## Summary

- advertise ACP audio prompt support
- map ACP base64 audio blocks to Codex App Server audio data URLs for normal prompts and steering
- reject audio explicitly when the selected Codex model does not advertise the audio input modality

## Why

The adapter currently returns `null` for every ACP audio block, silently removing user input. Codex App Server already accepts `UserInput::Audio` data URLs, and model metadata exposes whether audio input is supported. This change preserves supported audio end to end while preventing unsupported models from replacing it with an omission placeholder.

## Verification

- 37 test files passed, 331 tests passed
- `npm run typecheck`
- `npm run build`
- real authenticated text ACP smoke test completed with `end_turn`
- a real ACP audio probe reached App Server, which replaced it with an unsupported-audio omission for the selected model
- live model catalog check on Codex 0.145.0 confirmed all 7 current subscription models advertise only text/image, so the new unsupported-model guard is required today

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -934,12 +934,16 @@ function buildPromptItems(prompt: acp.ContentBlock[]): UserInput[] {
                 return {type: "text", text: `${link}\n${context}`, text_elements: []};
             }
             case "audio":
-                return null;
+                return {type: "audio", url: audioDataUrl(block)};
         }
     }).filter((block): block is UserInput => block !== null);
 }
 
 function imageDataUrl(block: acp.ContentBlock & { type: "image" }): string {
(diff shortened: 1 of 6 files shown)


shreyas-shinde:
Maintainer action needed: this PR is mergeable, but its GitHub Actions run is still blocked as action_required because the workflow has not been approved for this fork. Could a maintainer approve the run and review the ACP audio-content pass-through change? The implementation preserves negotiated capabilities and only forwards audio when the selected runtime advertises audio support; unsupported audio is rejected explicitly rather than silently dropped.

PR #348, opened 2026-07-31
Add shared App Server live-peer mode

moshthepitt:
## Why

ACP client should join Codex thread already open in stock TUI.

```text
stock Codex TUI
        |
same App Server + same thread
        |
ACP client
```

No shadow session. No TUI scraping. Codex owns transcript.

## What

- `CODEX_APP_SERVER_URL` attaches to Unix socket or WebSocket.
- Opt-in `_meta.codex.livePeer.version = 1`.
- Keep event, approval, elicitation handlers alive outside ACP-owned prompts.
- Emit terminal user messages and exact turn start/finish events.
- Pass `clientUserMessageId` through prompt and steering; echo it back.
- Frame authoritative history replay by turn.
- Report active turn on load/resume.
- Closing ACP closes its socket. Never shared App Server.

Capability advertisement:

```json
{
  "codex": {
    "livePeer": {
      "version": 1,
      "ambientEvents": true,
      "interactions": true,
      "userMessages": true,
      "clientUserMessageIds": true,
      "turnLifecycle": true,
      "historyReplay": true
    }
  }
}
```

No live-peer request: old behavior.

## Why replay framing

Transport reconnect alone lies. Client can miss completed turn.

Now load emits:

```text
history started
  turn started
  items
  turn finished
history finished
```

Client can baseline old history, reconcile missed turns, dedupe by native IDs.

## Proof

- Full suite: 340 passed, 28 skipped.
- Typecheck: pass.
- Build: pass.
- Stock Codex CLI 0.146.0:
  - bare TUI and ACP share one thread;
  - both directions render;
  - steering works;
  - approvals still work;
  - history load and runtime capabilities pass.

No Codex version pin. Client gates on negotiated capability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,41 @@ Use [OpenAI Codex](https://github.com/openai/codex) from [Agent Client Protocol]
 
 `codex-acp` is a stdio ACP agent server. It starts the Codex App Server, translates ACP requests into Codex operations, and maps Codex events back into the client.
 
+## Trunkline live-peer mode
+
+This downstream branch can connect to an App Server that is already shared with
+the stock Codex TUI:
+
+```bash
+CODEX_APP_SERVER_URL=unix:// codex-acp
+```
+
+`unix://PATH`, `ws://`, and `wss://` endpoints are also supported.
+
+Live-peer behavior is opt-in through ACP initialization metadata:
+
+```json
+{
+  "_meta": {
+    "codex": {
+      "livePeer": {
+        "version": 1
+      }
+    }
+  }
+}
+```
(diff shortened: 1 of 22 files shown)


robinfai:
Heads-up for a future rebase: #382 passes excludeTurns: true on the standard ACP resume/load paths. The live-peer resume path in this PR currently derives activeTurnId from response.thread.turns, so after rebasing it will need to omit excludeTurns for live-peer resume or query the active turn separately. session/load remains unaffected because it uses the subsequent thread/read response.

PR #353, opened 2026-08-02
fix: report full per-prompt token usage

zfy0701:
## Summary

- derive `PromptResponse.usage` from the cumulative token delta across the complete ACP prompt
- keep `usage_update.used` and quota metadata based on the latest response for context-status compatibility
- cover completed, cancelled, multi-update, and subsequent-prompt usage behavior

## Why

Codex app-server reports both the latest model response usage and cumulative session usage. The adapter currently returns only the latest response, so tool-heavy prompts with multiple model calls substantially under-report their per-turn usage.

The ACP maintainers' direction in [claude-agent-acp#390](https://github.com/agentclientprotocol/claude-agent-acp/issues/390#issuecomment-4183798512) is that `PromptResponse.usage` is per-turn. Taking the cumulative total at prompt completion minus its value at prompt start preserves that contract without double-counting across prompts.

## Validation

- `npm test` — 334 passed, 28 skipped
- `npm run typecheck`
- `npm run build`

Created by Codex . GPT-5.6 Sol

### DIFF
--- a/src/CodexAcpServer.ts
+++ b/src/CodexAcpServer.ts
@@ -35,7 +35,7 @@ import {
     REASONING_EFFORT_CONFIG_ID,
 } from "./ModelConfigOption";
 import type {TokenCount} from "./TokenCount";
-import {toPromptUsage} from "./TokenCount";
+import {subtractTokenCounts, toPromptUsage} from "./TokenCount";
 import {CodexCommands} from "./CodexCommands";
 import {SteeringQueue} from "./SteeringQueue";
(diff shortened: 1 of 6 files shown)


neubig:
@AlexandrSuhinin @ishulgin @benbrandt Hi all, we encountered this issue and it's hard to get proper costs out of the codex acp. This PR and https://github.com/agentclientprotocol/codex-acp/pull/367 both seem like attempts to solve the issue, would you be able to take a look?

PR #361, opened 2026-08-03
fix: propagate workspace write network access

jamesjlopez:
## Summary

`CODEX_CONFIG` can now enable network access for workspace-write turns. Previously, the adapter sent `networkAccess: false` on every prompt and overrode the configured setting. The merge accepts only boolean values, preserves writable roots, and leaves read-only and full-access policies unchanged.

Related: block/buzz#3031

## Testing

- `npm test` (345 passed, 28 skipped)
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/readme-dev.md
+++ b/readme-dev.md
@@ -13,6 +13,19 @@ Set `CODEX_PATH` to run a different Codex binary; versions other than the one sp
 - `NO_BROWSER` - hide browser-based ChatGPT auth when set.
 - `APP_SERVER_LOGS` - directory for adapter logs.
 
+`CODEX_CONFIG` can enable network access for the `agent` (workspace-write) mode:
+
+```json
+{
+  "sandbox_workspace_write": {
+    "network_access": true
+  }
+}
+```
+
+Only a boolean value overrides the workspace-write default of `false`. This setting
+does not change the `read-only` or `agent-full-access` policies.
+
 ### Quick start
 
 #### Develop on Windows?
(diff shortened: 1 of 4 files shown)


BradGroux:
Independent production-path validation from Buzz Desktop 0.5.7 confirms this PR addresses the load-bearing defect, and that the current published adapter remains affected.

### Validation from the affected path

- A managed Codex agent received its inbound Buzz mention and started normally.
- Buzz supplied `CODEX_CONFIG.sandbox_workspace_write.network_access=true`, while the resulting Codex child reported `CODEX_SANDBOX_NETWORK_DISABLED=1`.
- A read-only Buzz CLI command inside the child failed three out of three times with exit code 2 and the macOS `getaddrinfo`/`EAI_NONAME`-shaped error.
- An unsandboxed request to the same hosted relay in the same minute resolved, completed TLS, and returned HTTP/2 403, ruling out host DNS or relay reachability.
- The resulting permission request was rejected by Buzz's unattended fail-closed policy, so no reply could be published.

I also inspected the current npm release, `@agentclientprotocol/codex-acp` 1.1.14 (`gitHead` `5faefec5d55ded33c54b68ffec93def4f6c547f5`). Its bundle still hardcodes `networkAccess: false` for Agent mode and only merges `writable_roots`; upgrading from the affected 1.1.7 installation to 1.1.14 does not fix this path.

The patch here is correctly scoped from static review: it applies the configured value after additional roots are merged, accepts only a real boolean, changes only `workspaceWrite`, preserves the default false value, and leaves `dangerFullAccess` unchanged. The added tests cover enabled, absent, invalid, explicit false, preserved roots, and full-access behavior.

I did not rerun this branch's test suite locally; the evidence above is an independent production reproduction plus review of the complete diff. Once this lands, a prompt npm patch release would let Buzz raise its minimum supported adapter version and stop treating affected adapters as healthy.

PR #371, opened 2026-08-06
feat: expose provider-neutral ACP goal extension

nikita-ashihmin:
## Summary

- replace provider-specific `_meta.codex.goal` with a shared `_meta.goal` session-state extension
- advertise the neutral `_session/goal` control method and supported `pause` / `clear` actions
- keep goal activity independent from ACP prompt activity, so a persistent active goal does not occupy the session's prompt slot while Codex is quiescent
- retain `_codex/session/goal_control` only as an unadvertised compatibility alias
- document the extension as a candidate shape for a future first-class ACP goal API

## Architecture

A goal is persistent **session state**, not the lifecycle of one `session/prompt` request. `goal.status === "active"` says that the objective may drive more work; it does not say that a prompt is executing now.

This keeps two independent axes:

1. **Goal state** is published asynchronously as session metadata and may survive multiple prompt turns.
2. **Prompt activity** follows the backend's actual running/quiescent boundaries. While a turn runs, clients use advertised steering or prompt queueing. Once the backend is quiescent, an ordinary `session/prompt` is valid even if the goal remains active.

The previous version of this PR kept one ACP prompt open until the goal became terminal. That conflated these axes and could monopolize a client's single active prompt slot across idle gaps. The corrective commit removes that lifecycle state machine and restores completion at the current Codex turn's quiescent boundary.

The metadata is deliberately provider-neutral:

```json
{
  "_meta": {
    "goal": {
      "version": 1,
      "controlMethod": "_session/goal",
      "actions": ["pause", "clear"]
    }
  }
}
```

State is emitted as `session_info_update._meta.goal`; `null` clears it. The common snapshot vocabulary supports `objective`, common statuses, timestamps in Unix milliseconds, budgets/usage, iteration count, and a continuation reason. Each implementation emits only fields it actually knows. This shape is intentionally close to a future typed ACP capability/update so promotion out of `_meta` would not require a provider-specific migration.

Control actions are capability-negotiated. The shared vocabulary reserves `set`, `pause`, `resume`, and `clear`, but an agent advertises only the subset it can implement with correct semantics.

## Codex mapping

Codex app-server `thread/goal/updated` and `thread/goal/cleared` notifications are adapted at the ACP boundary:

- the objective, budget, token use, and elapsed time are copied into the common snapshot
- `usageLimited` and `budgetLimited` map to the common `limited` status
- app-server second timestamps are normalized to Unix milliseconds
- clear emits `_meta.goal: null`

The mapping lives behind a dedicated goal-extension contract and a provider adapter. Other Codex metadata remains under `_meta.codex`; goal state does not. `/goal` remains the user-facing route for setting/resuming goals. The control capability advertises only `pause` and `clear`, which Codex can perform independently.

For compatibility, `_codex/session/goal_control` remains accepted but is no longer advertised. New clients discover `_session/goal` from the capability.

The companion Claude implementation is agentclientprotocol/claude-agent-acp#964.

## Verification

- `npm test` — 344 passed, 28 skipped
- `npm run build`
- `npm run typecheck`


src/GoalExtension.ts
```diff
@@ -0,0 +1,33 @@
+import type {SessionId} from "@agentclientprotocol/sdk";
+
+export const GOAL_EXTENSION_VERSION = 1;
+export const GOAL_CONTROL_METHOD = "_session/goal";
+export const LEGACY_GOAL_CONTROL_METHOD = "_codex/session/goal_control";
+
+export type GoalControlAction = "set" | "pause" | "resume" | "clear";
```
EugeneTheDev: Let's reflect only actually supported actions here, i.e., pause and clear

src/GoalExtension.ts
```diff
@@ -0,0 +1,33 @@
...
+    updatedAt?: number;
+    tokenBudget?: number | null;
+    tokensUsed?: number;
+    timeUsedSeconds?: number;
+    controlMethod: typeof GOAL_CONTROL_METHOD;
+}
+
+export type GoalControlRequest = {
+    sessionId: SessionId;
+    action: "pause" | "clear";
```
EugeneTheDev: Let's reuse `GoalControlAction` for consistency

src/CodexAcpServer.ts
```diff
@@ -249,6 +251,11 @@ export class CodexAcpServer {
                 steering: {
                     supported: true,
                 },
+                goal: {
+                    version: GOAL_EXTENSION_VERSION,
+                    controlMethod: GOAL_CONTROL_METHOD,
+                    actions: ["pause", "clear"],
```
EugeneTheDev: The same, let's reuse `GoalControlAction`

EugeneTheDev (approved):

src/GoalExtension.ts
nikita-ashihmin: Fixed in 8c5c3b5: the action type is now derived from the actually supported pause/clear list.

src/GoalExtension.ts
nikita-ashihmin: Fixed in 8c5c3b5: GoalControlRequest.action now reuses GoalControlAction.

src/CodexAcpServer.ts
nikita-ashihmin: Fixed in 8c5c3b5: initialize metadata now uses the shared GOAL_CONTROL_ACTIONS value typed by GoalControlAction.

PR #373, opened 2026-08-06
Expose provider-neutral goal lifecycle controls

nikita-ashihmin:
## Summary
- publish session-scoped goal snapshots through the provider-neutral ACP goal extension
- keep goal lifecycle independent from prompt lifecycle and preserve goal state across session history
- advertise and handle pause, resume, and clear controls, including resuming a paused goal as active

## Validation
- npm run typecheck
- npm test (344 passed, 28 skipped)
- npm run build
- regression test reproduced the missing active status transition before the fix


nikita-ashihmin:
Superseded by #374. The provider-neutral goal extension landed independently in #371, so #374 is the clean one-commit resume fix on current main.

PR #379, opened 2026-08-09
feat: persist ACP system prompts as developer instructions

cmyk:
## Summary

ACP clients such as Buzz need a way to install stable managed-agent instructions once instead of repeating them inside every ordinary user prompt. `codex-acp` currently has no advertised persistent-system-prompt capability, so clients cannot safely select that behavior.

This PR adds the Codex side of the negotiated contract:

- advertise `_meta.capabilities.persistentSystemPrompt: true` during initialization;
- accept `_meta.systemPrompt` on new, resumed, and loaded ACP sessions;
- normalize and validate the value before any thread/session side effect;
- pass valid content to Codex as `developerInstructions`;
- reject malformed values and prompts larger than 256 KiB;
- omit developer instructions when the value is absent or blank;
- cover new, resume, and load paths plus exact validation boundaries.

This keeps static client-owned instructions in Codex's developer-instruction layer while leaving ordinary user turns free of repeated base/system material.

## Companion work

**Hard dependency for the accepted Buzz path:** the companion Buzz contribution negotiates this capability, sends its assembled static context once during `session/new`, and suppresses legacy repetition on ordinary turns:

https://github.com/block/buzz/compare/main...Peakhunter:buzz:fix/acp-persistent-system-prompt?expand=1

The Buzz publication head is `1db51a13a60e09f0876c41e163fc082453c45456`. Its source tree is byte-identical to the live-tested candidate head.

## Related work

- Buzz issue [#3242](https://github.com/block/buzz/issues/3242) tracks repeated full-context injection. This coordinated change addresses its static Base/System portion for adapters that advertise persistent prompt support; dynamic thread-context deduplication remains separate.
- Buzz issue [#5342](https://github.com/block/buzz/issues/5342) tracks loss of channel-session continuity after a Buzz restart. Live acceptance reproduced it, but the failure occurs because Buzz does not persist and resume its channel-to-session mapping; this adapter already supports `session/resume` and `session/load` when a client invokes them.
- Buzz PR [#4183](https://github.com/block/buzz/pull/4183) addresses unsupported protocol-v1 adapters by sending standing context once through the legacy user-message path. The coordinated PRs here instead negotiate developer/system instruction transport with the adapter, preserving instruction authority.

## Scope

This change owns persistent developer instructions only. It does not persist a client's channel-to-ACP-session mapping across client application restarts, hide Codex sessions from ChatGPT, alter worker-pool sizing, deduplicate dynamic thread context, or change sandboxed outbound network/delivery policy.

## Validation

Automated verification at exact head `1cf677dbe2801c4dbab678fc9e479a4a7fd87aa6`:

- **363 tests passed**;
- **28 E2E tests skipped**;
- TypeScript typecheck passed;
- rebuilt entry SHA-256 matched the previously recorded candidate exactly: `11120fdbfb9de40e3ad42a800f25b89b95a4719d082400acd1dbd65d50712ff2`.

Supervised macOS acceptance with the coordinated Buzz candidate established:

- initialization advertised `persistentSystemPrompt: true`;
- two ordinary turns continued one main Codex session;
- static Buzz instructions appeared once in initial developer instructions;
- the static material appeared zero times in both ordinary user-prompt records;
- after the client application restarted, the newly created session received the static instructions once and its ordinary prompt remained free of repeated static material.

The restart test also confirmed an existing client-side limitation: Buzz currently creates a new ACP/Codex session after a Desktop restart because its channel-to-session mapping is process-local. That lifecycle behavior is intentionally outside this adapter PR.

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -1075,7 +1075,7 @@ function readMetaAdditionalRoots(meta?: Record<string, unknown> | null): string[
         .filter(value => value.length > 0));
 }
 
-function readMetaSystemPrompt(meta?: Record<string, unknown> | null): string | undefined {
+export function readMetaSystemPrompt(meta?: Record<string, unknown> | null): string | undefined {
     const rawSystemPrompt = meta?.["systemPrompt"];
     if (rawSystemPrompt === undefined) {
         return undefined;
@@ -1083,13 +1083,13 @@ function readMetaSystemPrompt(meta?: Record<string, unknown> | null): string | u
     if (typeof rawSystemPrompt !== "string") {
         throw RequestError.invalidParams(undefined, "systemPrompt must be a string");
     }
+    if (new TextEncoder().encode(rawSystemPrompt).byteLength > 256 * 1024) {
+        throw RequestError.invalidParams(undefined, "systemPrompt must not exceed 262144 UTF-8 bytes");
+    }
 
     const systemPrompt = rawSystemPrompt.trim();
     if (systemPrompt.length === 0) {
         return undefined;
-    }
-    if (new TextEncoder().encode(systemPrompt).byteLength > 256 * 1024) {
-        throw RequestError.invalidParams(undefined, "systemPrompt must not exceed 262144 UTF-8 bytes");
     }
     return systemPrompt;
 }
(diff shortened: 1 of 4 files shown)


cmyk:
Companion Buzz PR is now submitted:

- block/buzz#5386 — https://github.com/block/buzz/pull/5386
- exact head: `1db51a13a60e09f0876c41e163fc082453c45456`

That PR negotiates `persistentSystemPrompt`, sends Buzz-owned static context once during ACP session creation, and omits it from ordinary turns. Together, these two PRs are the coordinated path used for the live acceptance reported above.

AmadeusK525:
I'm also interested in this feature, I find it quite surprising that this ACP server does not support receiving system prompts

PR #383, opened 2026-08-09
feat: expose typed session failures for AIR

nikita-ashihmin:
## Summary

- Negotiate the JetBrains AIR `sessionFailure` extension when the client advertises an integer version `>= 1` and the capability name.
- Return turn-terminal failures only on `PromptResponse._meta`; publish late, idle, and otherwise session-scoped failures through `session/update`.
- Keep legacy ACP cancellation and `RequestError` behavior for capability-off clients.
- Emit bounded fixed safe messages without provider details, stderr, paths, headers, tokens, or stack traces.

## Recovery contract

Failures use `_meta.jetbrains.air.sessionFailure` with a stable id, monotonic revision, phase, category, source, safe message, retryability, and optional real turn attribution. Turn failures use the completed app-server turn id. Session-scoped ids include a per-server epoch so an AIR replay tombstone cannot suppress a new outage after process restart.

AIR owns localized action presentation. Codex supplies typed category and retryability facts. A successful prompt clears only the exact failure revision that was active when that recovery attempt started; a newer failure is never cleared accidentally.

The event routing covers the boundaries that previously lost or misattributed errors: before turn start, after `turn/completed`, during plan approval, between approval and implementation `turnStart`, during implementation, after prompt return, process exit, and local slash commands that complete without creating a turn.

## Validation

- 394 passed, 28 skipped.
- Build passed.
- Typecheck passed.
- Diff check passed.
- In-memory NDJSON tests cover initialize negotiation, terminal `PromptResponse` metadata, session-scoped updates, process exit, capability-off behavior, restart-safe ids, and plan/slash-command races.

### DIFF
--- a/src/AirExtension.ts
+++ b/src/AirExtension.ts
@@ -1,0 +1,14 @@
+/**
+ * Wire names for the versioned JetBrains AIR ACP extension.
+ *
+ * `jetbrains.air` is a protocol namespace, not user-facing branding: `jetbrains`
+ * owns the non-standard contract and `air` identifies the client that defines its
+ * rendering semantics. Keeping both levels prevents unrelated JetBrains ACP clients
+ * from accidentally interpreting or colliding with this metadata.
+ */
+export const JETBRAINS_META_KEY = "jetbrains";
+export const AIR_META_KEY = "air";
+export const AIR_EXTENSION_VERSION_KEY = "version";
+export const AIR_EXTENSION_CAPABILITIES_KEY = "capabilities";
+export const AIR_SESSION_FAILURE_KEY = "sessionFailure";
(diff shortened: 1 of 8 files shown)


Ololoshechkin (approved):
@nikita-ashihmin please check my comments in related PR https://github.com/agentclientprotocol/claude-agent-acp/pull/979#pullrequestreview-4895541988 -- same points are applicable here as well

nikita-ashihmin:
Applied both comments from the related Claude PR in 2ccc37d: all AIR extension wire keys are centralized and shared by negotiation/emission, and the code now documents why `jetbrains.air` is a two-level protocol namespace (vendor + rendering contract), not user-facing branding.

PR #390, opened 2026-08-11
feat: identify skill loads in _meta.codex.skill

nikita-ashihmin:
## Problem

Codex has no skill tool. It loads a skill by reading that skill's `SKILL.md`, which reaches the client as an ordinary `read` command action — so a skill load is indistinguishable from any other file read, and clients render it as `Read file '/…/skills/commits/SKILL.md'`.

A client can guess from the path's shape (`<…>/skills/<name>/SKILL.md`), but that guess is wrong whenever a skill's directory name differs from its registered name, and it misses skills that don't live under a `skills` directory at all.

## Change

`skills/list` already returns each skill's `name` and absolute `path`, and `refreshSkills` already calls it — the response was just discarded. This indexes it by `SKILL.md` path, and when a `read` command action hits a known skill file, the tool call carries:

```json
{ "_meta": { "codex": { "skill": "commits", "skillPath": "/…/.agents/skills/commits/SKILL.md" } } }
```

The title also becomes `Read skill 'commits'` instead of the raw path, and `locations` still points at the file so existing navigation keeps working.

Because the name comes from Codex's own registry rather than from the path, it stays authoritative in both cases above.

The lookup is threaded in as an optional parameter, so history replay — which has no live registry — leaves such reads as ordinary file reads instead of guessing. No existing snapshot changes.

## Client side

This mirrors `_meta.claudeCode.skill`/`skillPath` in claude-agent-acp. JetBrains AIR consumes both to render a skill load as its own block, with the skill name linking to `SKILL.md`.

## Testing

- `npm run typecheck` clean
- `npm test` — 404 pass, 28 skipped, no snapshot churn
- New: three cases in `command-action-events.test.ts` (known skill file → named load; unknown file → plain read; no registry → plain read) and one in `CodexAcpClient.test.ts` (`skills/list` populates the index, unrelated paths miss)

### DIFF
--- a/src/CodexAcpClient.ts
+++ b/src/CodexAcpClient.ts
@@ -51,7 +51,7 @@ import packageJson from "../package.json";
 import type {AuthenticationStatusResponse} from "./AcpExtensions";
 import {createCodexCollaborationMode} from "./CollaborationModeConfig";
 import type {ModeKind} from "./app-server/ModeKind";
-import {arePathBasenamesEqual, arePathsEqual, isAbsolutePathLike} from "./PathUtils";
+import {arePathBasenamesEqual, arePathsEqual, isAbsolutePathLike, normalizePathForComparison} from "./PathUtils";
 
 /**
  * Well-known provider id for the client-configurable custom LLM gateway.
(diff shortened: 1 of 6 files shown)


nikita-ashihmin:
Closing: this belongs on the client side instead. A skill load is recognizable from the read itself — the `SKILL.md` filename plus its containing directory as the skill name — so the client can do it for any agent without codex-acp carrying skill metadata. Keeping the agent-side special case only where an agent genuinely has a skill tool (Claude's `Skill`).

PR #394, opened 2026-08-12
fix: persist ACP model and reasoning effort across turns and resume

### LINKED ISSUE #336: Session "set reasoning_effort" applies to the next turn only, then reverts to the config default
**Setup:** codex-acp driven as a persistent named session over ACP (client: [acpx](https://github.com/openclaw/acpx)). Tested on codex-acp 1.1.4; I found no related fix in the 1.1.5–1.1.7 changelogs and no existing issue matching this.

**Reproduction:**

1. Create a session, `set model gpt-5.6-sol`, `set reasoning_effort xhigh` — the set acks (`config set: reasoning_effort=xhigh (5 options)`).
2. Send a prompt. The codex rollout jsonl (`~/.codex/sessions/.../rollout-*.jsonl`) records `"effort":"xhigh"` in that turn's `turn_context` — the set applied.
3. Send further prompts on the same session **without** re-issuing the set. Each later `turn_context` records `"effort":"high"` (the codex config default) — the session setting silently reverted.

**Observed twice independently:** a three-turn session with turn 1 at `xhigh` and turns 2–3 at `high`; a longer multi-turn session with 29 `turn_context` entries at `high` vs 1 at `xhigh`.

**Expected:** session-level reasoning effort persists for the session's lifetime, like `set model` does. The model set through the same path sticks across all turns, which is why this looks like adapter-side session-config handling rather than the client dropping it — but happy to re-test against a newer version or provide full rollout excerpts if useful.

**Workaround:** re-issue `set reasoning_effort` immediately before every prompt.

---
<sub><em>**[@superbiche](https://github.com/superbiche)** · user · drafted with Claude Fable 5, reviewed before posting; the voice and decisions are mine.</em></sub>

### LINKED ISSUE #343: session/load returns the config.toml model/effort instead of the thread's, because thread/resume is always sent a modelProvider
**Setup:** codex-acp driven over ACP by [Kepler](https://github.com/gitkraken/kepler) (Electron client, spawns the wrapper via `dist/index.js` and resumes stored sessions with `session/load`). Reproduced on codex-acp 1.1.4 and 1.1.7, against `codex-cli 0.144.6` app-server.

## Problem

`session/load` returns the model and reasoning effort from `~/.codex/config.toml` instead of the ones persisted on the thread. Any in-session model/effort switch is silently lost on resume.

The cause is the `modelProvider` field that `loadSession`/`resumeSession` always pass to `thread/resume`:

```js
// dist/index.js (1.1.7:26277) — same in 1.1.4:26044
const response = await this.codexClient.threadResume({
  config: await this.createSessionConfig(...),
  cwd: request.cwd,
  modelProvider: await this.getResumeModelProvider(),  // never absent
  threadId: request.sessionId,
});
```

`getResumeModelProvider()` is `getCurrentModelProvider() ?? "openai"`, so the field is always populated even with no gateway configured. Passing it makes the app-server discard the thread's stored `model`/`reasoningEffort` and re-resolve them from config.

## Reproduction

1. Create a session and switch it to a model/effort that differs from `~/.codex/config.toml`. Mine has `model = "gpt-5.5"`, `model_reasoning_effort = "xhigh"`; the session was set to `gpt-5.6-sol` / `max`.
2. Send a prompt. The rollout jsonl records the switch: `"model":"gpt-5.6-sol"`, `"effort":"max"` in `turn_context`.
3. Resume that thread with `session/load`.

Observed on both versions:

```
currentModelId= gpt-5.5[xhigh]
  model            => "gpt-5.5"
  reasoning_effort => "xhigh"
```

Expected `gpt-5.6-sol` / `max` (what the thread holds).

## Isolating it to `modelProvider`

Calling `thread/resume` directly on the app-server for the same thread, varying only which params are sent:

| params sent | `model` | `reasoningEffort` |
| --- | --- | --- |
| `threadId`, `cwd` | `gpt-5.6-sol` | `max` |
| + `config` | `gpt-5.6-sol` | `max` |
| + `modelProvider: "openai"` | `gpt-5.5` | `xhigh` |
| + `config` + `modelProvider` (what codex-acp sends) | `gpt-5.5` | `xhigh` |

`config` is irrelevant; `modelProvider` alone flips it. The values it falls back to are exactly the ones in `config.toml`.

## Expected

Resuming a thread should restore the model and reasoning effort stored on that thread. Either don't send `modelProvider` on `thread/resume` when no gateway is configured, or have the app-server keep the thread's model when a provider is passed. If the app-server behaviour is the intended contract, this is arguably an `openai/codex` bug and I'm happy to move it there.

## Impact

The client cannot recover the picks on its own: after a client restart there is no live session left to read them from, and the resume response is the only source of truth. Users see the correct model for a moment (from the client's own cache), then it flips to the config default as the load response lands.

Possibly the same root cause as #336, which reports effort reverting between turns of a live session — that one survives without a resume, so it may be a separate path. Filing this separately since the trigger here is isolated to `session/load`.

superbiche:
This refreshes @gilbertl's #337 onto current `main` (v1.7.0), preserving the original commits and authorship while resolving the conflicts. The scope is intentionally narrower than the earlier revision: model and reasoning-effort persistence only; agent-mode behavior remains identical to upstream `main`.

What changes:

- `model` and `reasoning_effort` configuration updates are persisted with `thread/settings/update` and acknowledged only after Codex accepts them. A setting therefore remains effective for subsequent turns and process reconnects until changed.
- Resume, load, and fork no longer invent `modelProvider: "openai"` when no provider is configured. Omitting it lets the Codex thread retain its stored model and reasoning effort.
- Regression coverage exercises persistence across an adapter restart and session load.

Validation on the final candidate (`0275316`):

- `npm run typecheck`: pass
- `npm test`: 486 passed, 27 skipped
- `npm run bundle:all`: all six platform bundles pass
- `npm run build`: pass
- Live ChatGPT-auth reconnect oracle: selected effort changed `high` → `low`; after two independent adapter restarts, both resume and load reported `low`
- Independent deep review: three rounds, final `PASS`; agent-mode persistence was removed during review instead of adding concurrency complexity unrelated to this fix

Known boundary: concurrent ACP callers can still race a setting update against a prompt already in flight. The deployed `acpx` flow is sequential and waits for the setting acknowledgement before sending the next prompt.

Transport-level effective-config receipts need an authoritative app-server receipt, and per-invocation effort for one-shot `acpx exec` belongs in `acpx`; neither is bundled into this adapter fix.

Fixes #336.
Fixes #343.

---
<sub><em>**[@superbiche](https://github.com/superbiche)** · user · drafted with Codex (GPT-5).</em></sub>


superbiche:
This conflicted branch has been superseded by #457, rebased onto current `main` / v1.8.0 with the persistence, prompt-serialization, resume/load/fork, and regression-test updates intact.

I left this PR open rather than rewriting its published history; maintainer review should continue on #457.

---
<sub><em>**[@superbiche](https://github.com/superbiche)** · user · drafted with Codex (GPT-5).</em></sub>

superbiche:
Closing in favor of #457

PR #404, opened 2026-08-14
feat: switch providers for loaded Codex sessions

serjic:
### Problem

Codex fixes `modelProvider` in the app-server thread runtime. Updating the adapter's Providers API state therefore affected newly created threads, while already loaded ACP sessions continued sending turns through the old backend.

The app-server does not currently expose an API that replaces a loaded thread's provider/config while preserving its ID.

### Solution

- Serialize provider updates and wait for active prompts to complete.
- Restart the adapter-owned Codex app-server between turns.
- Reinitialize the replacement client and resume every loaded thread with its existing session ID, cwd, additional directories, and MCP servers.
- Preserve startup/native provider state and restore it on `providers/disable`.
- Block session operations only while the coordinated restart is in progress.
- Attempt every thread resume even if one fails, report aggregate failures, and allow a later provider update to recover.
- Add PID and provider/restart routing details to opt-in app-server logs without exposing provider headers.

### Tests

Added coverage for:

- native behavior when Providers API is unused;
- gateway config passed to `thread/start`;
- multiple loaded sessions across native → proxy A → proxy B → native;
- all-session resume after app-server restart;
- partial resume failure and successful subsequent recovery;
- startup proxy restoration and legacy gateway auth interoperability.

Verification:

- provider suite: 14 passed
- `npm run typecheck`
- `npm run build`
- `git diff --check`

The repository's full suite still has the pre-existing approval/elicitation/session-close fixture mismatches; the provider suite and static/build checks are green.

### DIFF
--- a/src/CodexAcpServer.ts
+++ b/src/CodexAcpServer.ts
@@ -7,12 +7,14 @@ import {type CodexAuthRequest, getCodexAuthMethods, isCodexAuthRequest} from "./
 import {clientSupportsUrlElicitation} from "./ElicitationCapabilities";
 import {
     CodexAcpClient,
+    type JsonObject,
     OPENAI_PROVIDER_ID,
     type SessionMetadata,
     type SessionMetadataWithThread,
     type UrlElicitationRequester
 } from "./CodexAcpClient";
-import type {McpStartupResult} from "./CodexAppServerClient";
+import {CodexAppServerClient, type McpStartupResult} from "./CodexAppServerClient";
+import {type CodexConnection, startCodexConnection} from "./CodexJsonRpcConnection";
 import {type AcpClientConnection, ACPSessionConnection, type UpdateSessionEvent} from "./ACPSessionConnection";
 import type {InputModality, ReasoningEffort} from "./app-server";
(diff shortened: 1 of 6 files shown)


src/CodexAcpServer.ts
```diff
@@ -194,20 +196,22 @@ interface ActivePrompt {
     complete: () => void;
 }
 
+export type RestartCodexClient = () => Promise<CodexAcpClient>;
```
EugeneTheDev: Do we really need an extra type here? It's used only in a couple of places, so we're not saving much by introducing a type alias. IMHO the raw type reads better

src/index.ts
```diff
@@ -109,10 +113,38 @@ function startAcpServer() {
 
     const acpJsonStream = createJsonStream(process.stdin, process.stdout);
 
+    async function restartCodexClient(): Promise<CodexAcpClient> {
```
EugeneTheDev: Can't we move this function to the `CodexAcpServer` instead of passing it to the server class as lambda? IMHO might look a bit cleaner, since it's a server's concern to restart a codex client

EugeneTheDev (approved):

PR #405, opened 2026-08-14
feat: add ACP v1 permission presentation

nikita-ashihmin:
## Summary
- emit autonomous ACP v1 tool-call snapshots with minimal `_meta.permission` presentation
- map only authoritative Codex decisions and keep exact provider payloads adapter-side
- distinguish explicit reject from cancellation and remove legacy rich permission metadata

## Tests
- `npm test` (405 passed, 28 skipped)
- `npm run typecheck`
- `npm run build`

## Known rollout constraint
`@openai/codex` 0.147 generated approval params do not declare `availableDecisions`. The adapter validates the runtime field and fails closed with `cancel` when it is absent or malformed rather than inventing choices.


src/permissions/command-decision-contract.ts
```diff
@@ -0,0 +1,102 @@
...
+        const allowAmendment = params.proposedNetworkPolicyAmendments?.find(amendment => amendment.action === "allow");
+        if (allowAmendment) {
+            decisions.push({applyNetworkPolicyAmendment: {network_policy_amendment: allowAmendment}});
+        }
+        decisions.push("cancel");
+        return decisions;
+    }
+    if (params.additionalPermissions) return ["accept", "cancel"];
+    const decisions: CommandExecutionApprovalDecision[] = ["accept"];
+    if (params.proposedExecpolicyAmendment) {
```
Ololoshechkin: ` && params.proposedExecpolicyAmendment.length > 0` from the original `CodexApprovalHandler.ts` is lost, so an empty amendment array becomes an offerable allow_always decision with an empty prefix.

src/permissions/command-decision-contract.ts
```diff
@@ -0,0 +1,102 @@
...
+        const decisions: CommandExecutionApprovalDecision[] = ["accept", "acceptForSession"];
+        const allowAmendment = params.proposedNetworkPolicyAmendments?.find(amendment => amendment.action === "allow");
+        if (allowAmendment) {
+            decisions.push({applyNetworkPolicyAmendment: {network_policy_amendment: allowAmendment}});
+        }
+        decisions.push("cancel");
+        return decisions;
+    }
+    if (params.additionalPermissions) return ["accept", "cancel"];
+    const decisions: CommandExecutionApprovalDecision[] = ["accept"];
```
Ololoshechkin: The legacy fallback for plain commands drops both acceptForSession ("don't ask again this session") and decline (turn-continuing rejection) that the old handler always offered, and the network branch keeps only the first allow amendment, silently discarding all deny ("block this host") amendments.

Ololoshechkin (approved):

PR #415, opened 2026-08-19
feat: estimate cost for usage updates

gilgm12:
## Summary
- fetch OpenAI model pricing and attach estimated USD cost to ACP usage_update events
- support both new and resumed/loaded sessions, baselining resumed cumulative token totals so Superlite can add persisted prior cost without double-counting
- add focused tracker coverage for cumulative deltas and resume baselining

## Verification
- npm exec vitest run src/__tests__/SessionCostTracker.test.ts src/__tests__/CodexACPAgent/token-usage-events.test.ts
- npm run typecheck
- npm run build


gilgm12:
@benbrandt let me know if there's anything else you need from me on this.

Issue #215, opened 2026-06-23
Support ACP session metadata pass-through for Codex threadStart/config


lichao2014:
﻿### Summary

`codex-acp` currently accepts a narrow set of ACP session inputs and maps them into Codex app-server state. For integrations that need per-session behavior, there does not seem to be a documented way to forward metadata such as `systemPrompt`, developer instructions, or arbitrary Codex `threadStart.config` overrides.

### Use case

ACP clients such as OpenClaw/acp-harness need to forward their runtime prompt and session options to external ACP agents.

For example, Claude ACP can consume `_meta.systemPrompt`, but `codex-acp` currently appears to ignore it. This makes it hard for ACP clients to provide equivalent behavior across agents.

### Current behavior

From `src/CodexAcpClient.ts` on `main`:

- `newSession()` calls `codexClient.threadStart({ config, modelProvider, cwd })`.
- `createSessionConfig()` builds project trust roots, sandbox writable roots, MCP server config, and gateway provider config.
- `_meta.additionalRoots` is read as a fallback for `additionalDirectories`, then mapped into Codex skill extra roots.
- `_meta.systemPrompt` and other generic session options are not mapped into Codex app-server `threadStart` params or `config`.

### Proposed behavior

Please add a documented pass-through mechanism for ACP clients to provide Codex session/thread options.

One possible shape:

```ts
_meta: {
  codex?: {
    config?: Record<string, unknown>;
    threadStart?: {
      baseInstructions?: string;
      developerInstructions?: string;
      config?: Record<string, unknown>;
    };
  };

  // optional compatibility alias used by some ACP clients
  systemPrompt?: string | { append: string };
}
```

Possible mapping:

- `_meta.codex.config` merges into the existing `threadStart.config`.
- `_meta.codex.threadStart.baseInstructions` maps to Codex app-server `baseInstructions`.
- `_meta.codex.threadStart.developerInstructions` maps to Codex app-server `developerInstructions`.
- `_meta.systemPrompt` could either be ignored explicitly, documented as unsupported, or mapped as a compatibility alias to `developerInstructions` / `config.instructions`.

### Why this matters

ACP clients can already pass session metadata through `_meta`, and different ACP agents expose similar functionality differently. A Codex-specific pass-through would let clients integrate Codex ACP without patching or forking `codex-acp`.

It would also make skills/prompt/session behavior easier to align across ACP adapters while keeping Codex-specific behavior namespaced under `_meta.codex`.

[...]

PR #454, opened 2026-09-01
feat: support system prompt append

Shuryne:
## Summary

- advertise a versioned system prompt append extension during ACP initialization
- validate client-provided append metadata and map it to Codex developer instructions without replacing base instructions
- apply explicit append metadata across new, resume, load, and fork lifecycle paths while omitting absent values
- document the extension and add a real-Codex test-runner option

## Validation

- npm run typecheck
- npm run build
- 20 targeted system-prompt, initialization, and fork tests passed
- 495 additional tests passed and 26 were skipped when excluding the existing MCP config integration file
- real Codex rollout verified the appended text is emitted in a separate developer-role message alongside the existing base, skills, and permission instructions

## Local environment note

The repository-pinned Codex 0.148 platform executable was not runnable on this macOS host, so real-Codex verification used the installed Codex 0.151 binary. Under that substitution, three existing MCP config merge tests exited during app-server initialization; all other tests passed, and this change does not touch MCP configuration.

### DIFF
--- a/.claude/skills/run-codex/scripts/run-codex-test.ts
+++ b/.claude/skills/run-codex/scripts/run-codex-test.ts
@@ -17,12 +17,19 @@ import {CodexAcpServer} from "../../../../src/CodexAcpServer";
 import type {AgentSideConnection} from "@agentclientprotocol/sdk";
 
 // Parse command line arguments
-function parseArgs(): { prompt: string; cwd: string; output: string; json: boolean } {
+function parseArgs(): {
+    prompt: string;
+    cwd: string;
+    output: string;
+    json: boolean;
+    systemPromptAppend?: string;
+} {
     const args = process.argv.slice(2);
     let prompt = "";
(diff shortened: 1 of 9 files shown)


salman1993:
it'd be great to allow overriding or appending to the codex system prompt. claude-agent-acp allows this now: https://github.com/agentclientprotocol/claude-agent-acp/pull/91 @nikita-ashihmin

salman1993 (changes requested):
🤖  automated review:

- Support replacement via `baseInstructions`, alongside append via `developerInstructions`.
- Define load/resume/fork behavior, noting that Codex ignores overrides when resuming a running thread.
- Avoid silently replacing existing `developer_instructions`.
- Add a real-Codex behavior test for both modes.

The V2 API already supports both fields, so these should be incremental changes. see comments in this issue - https://github.com/agentclientprotocol/codex-acp/issues/215

PR #464, opened 2026-09-03
fix: hide browser ChatGPT auth on headless hosts and fail fast

OldKrab:
## Problem

Selecting **ChatGPT → Sign in** from OpenAIDE running as a headless systemd unit did nothing: ACP `authenticate` was sent and never answered. `authenticateWithChatGpt` calls `open(authUrl)` on the agent host and then awaits `account/login/completed`. With no display server there is no browser, so no loopback callback ever arrives and the request hangs silently.

## Fix

- `browserSignInAvailable(env, platform)`: false when `NO_BROWSER` is set, or on Linux without `DISPLAY`/`WAYLAND_DISPLAY`/WSL interop. `getCodexAuthMethods` uses it, so headless hosts advertise only `api-key` and `chat-gpt-device-code` (when the client supports URL elicitation).
- `authenticateWithChatGpt` rejects with an explicit error on such hosts instead of waiting forever.
- README / readme-dev updated.

## Tests

- `initialize.test.ts`: headless Linux hides `chat-gpt`, keeps `api-key` + `chat-gpt-device-code`; DISPLAY / WAYLAND / WSL / darwin / win32 keep it.
- `CodexAcpClient.test.ts`: `chat-gpt` on a headless host rejects and never calls `accountLogin`.
- `npm test`: 495 passed.

Made with [Cursor](https://cursor.com)


OldKrab:
Opened against the wrong repository by mistake; this belongs to the OldKrab fork. Apologies for the noise.

PR #497, opened 2026-09-11
fix: mark ripgrep no-match as completed

### LINKED ISSUE #468: Ripgrep no-match exit code 1 is translated into an ACP failed tool call
## Summary

`codex-acp` reports a ripgrep search that found nothing as a `failed` ACP tool call. Exit code 1 from ripgrep means "no matches were found." It is not an error. Codex marks every nonzero exit as a failure, and the adapter passes that status straight through.

ACP clients then give a successful negative search the same red failure treatment as a malformed regex, a missing path, a permission denial, or a process that never launched.

## Environment

- `@agentclientprotocol/codex-acp` 1.8.0
- Windows
- PowerShell
- Codex through ACP
- I hit this downstream in Codeg Desktop 0.30.2

## Reproduction

1. Start Codex through `codex-acp`.
2. Run a standalone search for a token that does not exist:

```powershell
rg -n "__definitely_absent_token__" README.md
```

3. The command writes nothing to stdout or stderr, and ripgrep exits 1.
4. Observe the ACP update:

```json
{
  "sessionUpdate": "tool_call_update",
  "status": "failed",
  "rawOutput": {
    "formatted_output": "",
    "exit_code": 1
  },
  "_meta": {
    "terminal_exit": {
      "exit_code": 1,
      "signal": null
    }
  }
}
```

## Repeated evidence

The same thing happened across several independent Codex ACP sessions. Three standalone `rg` searches exited 1 with empty output, and the adapter marked all three failed.

One compound command shows why you cannot simply remap every exit 1. An earlier command in the chain produced valid output, then a final no-match search set the shell's overall status.

Other transcripts cover malformed regexes, missing paths, and invalid Windows globs. Every one of those wrote a diagnostic. Those are real errors and should stay failures.

## Why it happens

Ripgrep documents these exit codes:

- `0`: a match was found
- `1`: no matches were found
- `2`: an error occurred

Codex classifies any nonzero shell exit as a failed command. openai/codex#8268 reports exactly this behavior upstream, and it is closed as not planned. That leaves the adapter as the practical place to fix it.

The adapter forwards the Codex item status directly:

- `src/CodexEventHandler.ts`, `completeCommandExecutionEvent`: `item.status === "completed" ? "completed" : "failed"`
- `src/CodexToolCallMapper.ts`, `createCommandExecutionCompleteUpdate`: the same mapping

Both paths keep `item.exitCode` and `item.aggregatedOutput`, and the opening command event carries the command text. That is enough to recognize a standalone `rg` no-match.

## Expected behavior

A standalone ripgrep search that exits 1 with no diagnostic should not go out as an ACP failure. Report it as a completed search that found nothing, and keep the real exit code in metadata.

If ACP has no neutral no-match state today, this mapping works:

- `status: "completed"`
- preserve `rawOutput.exit_code: 1`
- optionally add adapter metadata with `no_matches: true`

## Safety constraints

Please do not treat every exit code 1 as success. A fix needs to separate four cases:

1. standalone `rg` with matches, which is completed;
2. standalone `rg` with no matches and no diagnostics, which is completed with no match;
3. malformed regex, missing path, permission error, or launch error, which stay failures;
4. compound and pipelined commands, which keep the shell's semantics unless the final command can be classified safely.

Tests should cover the live event path and the history/replay mapping, so a reloaded session shows the same status as the live one.

## Impact

- ACP clients show misleading red failure states.
- Agents conclude that ripgrep or the environment is broken.
- Agents retry valid searches or fall back to slower file scanning.
- Tool-call status stops being a reliable signal when you review what an agent actually did.

kocaemre:
## Summary

- Treat standalone `rg`/`ripgrep` commands that exit 1 with no output as completed ACP tool calls instead of failures.
- Reuse the status mapping for both live command completion events and session-history replay.
- Keep diagnostics, exit code 2, and compound shell commands as failures.

Closes #468.

## Test Plan

- `npm test -- src/__tests__/CodexACPAgent/command-execution-status.test.ts`
- `npm run typecheck`
- `npm run build`
- `npm test`
- `git diff --check`

### DIFF
--- a/src/CodexEventHandler.ts
+++ b/src/CodexEventHandler.ts
@@ -39,6 +39,7 @@ import type {
     WarningNotification
 } from "./app-server/v2";
 import type { McpStartupCompleteEvent } from "./app-server/McpStartupCompleteEvent";
+import { commandExecutionAcpStatus } from "./CommandExecutionStatus";
 import {toTokenCount} from "./TokenCount";
 import {
     commandExecutionUsesTerminalOutput,
@@ -1009,7 +1010,7 @@ export class CodexEventHandler {
         const update: UpdateSessionEvent = {
             sessionUpdate: "tool_call_update",
             toolCallId: item.id,
-            status: item.status === "completed" ? "completed" : "failed",
+            status: commandExecutionAcpStatus(item),
             rawOutput: {
                 formatted_output: item.aggregatedOutput ?? "",
                 exit_code: item.exitCode
--- a/src/CodexToolCallMapper.ts
+++ b/src/CodexToolCallMapper.ts
@@ -4,6 +4,7 @@ import { readFile } from "node:fs/promises";
 import path from "node:path";
 import type { UpdateSessionEvent } from "./ACPSessionConnection";
 import { stripShellPrefix } from "./CommandUtils";
+import { commandExecutionAcpStatus } from "./CommandExecutionStatus";
 import type {
     FuzzyFileSearchSessionCompletedNotification,
     FuzzyFileSearchSessionUpdatedNotification
@@ -110,7 +111,7 @@ export function createCommandExecutionCompleteUpdate(
     const update: UpdateSessionEvent = {
         sessionUpdate: "tool_call_update",
         toolCallId: item.id,
-        status: item.status === "completed" ? "completed" : "failed",
+        status: commandExecutionAcpStatus(item),
         rawOutput: {
             formatted_output: item.aggregatedOutput ?? "",
             exit_code: item.exitCode,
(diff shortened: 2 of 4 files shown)


kocaemre:
I rebased this fix onto current `main`, but this PR's GitHub head stayed pinned to the old `f24cd47` even after the fork branch updated to `5e2be67` (`git ls-remote` and the fork branch API both show the new SHA while the PR API still reports the old SHA).

To avoid leaving a conflicting/stale PR open, I opened a fresh replacement PR with the same focused fix and current-base validation: #535.

Closing this one in favor of #535.

PR #515, opened 2026-09-14
feat: Add experimental ACP session compaction updates

benbrandt:
Implements the agent side of the proposed RFD, gated by the capability


nikita-ashihmin:
Hi we hacve added similar for jetbrains air we need to merge if use standard solution

benbrandt:
yes I tried to keep the old solution there if the client does not expose the new capability

benbrandt:
I have checked that this will only affect client that advertise the capability, so should be safe enough and AIR will be able to opt-in when ready

PR #517, opened 2026-09-15
feat: await requested MCP servers startup outcome on new session start

EugeneTheDev:
Currently, session/new starts a new thread and processes incoming requests immediately, without waiting for the requested MCP servers to be ready. Introduce await when starting/resuming/forking a session to make sure all MCPs are available before the first user request.


src/CodexAcpServer.ts
```diff
@@ -702,12 +702,25 @@ export class CodexAcpServer {
...
+                await pendingStartup.startup;
+            } catch (err) {
+                if (this.pendingMcpStartupSessions.get(sessionId) === pendingStartup) {
+                    this.pendingMcpStartupSessions.delete(sessionId);
+                }
+                throw err;
+            }
+            if (canPublishSessionUpdates) {
+                this.publishMcpStartupStatusAsync(sessionId);
+            }
```
Rizzen, line 723: As far as I understand - in case we have any MCP servers - we wait for their init always and for unbounded time. Two questions/concerns:
1. Maybe we want to make that wait optional or opt-out?
2. It might be better to introduce some waiting timeout here

What do you think?

Rizzen (changes requested):

PR #518, opened 2026-09-15
perf: derive file change reports from turn diffs

nikita-ashihmin:
## Summary
- derive the existing AIR `agentFileChangeReport` from Codex `turn/diff/updated` snapshots
- remove the hidden ephemeral fork and extra model turn
- preserve the version-1 wire contract and path normalization limits
- cover aggregated snapshots, renames, binary files, quoted paths, cancellation, and failures

## Validation
- `npm run typecheck`
- `npm test` (660 passed, 26 skipped)
- `npm run build`
- `git diff --check`


nikita-ashihmin:
Reviewed the latest multi-environment finding and found it non-actionable for this adapter. Codex 0.154 app-server `ThreadStartParams` and `TurnStartParams` expose no environment-selection field, and codex-acp does not call any environment-selection API; its turns use the single local environment created from `cwd` plus workspace roots. The multi-environment prefixing branch in core is therefore not reachable through this ACP request surface. No code change was made for that hypothetical mode.

nikita-ashihmin:
Reviewed the latest empty-aggregate finding against the pinned Codex 0.154 source. It is not reproducible: `emit_patch_end` records `had_unified_diff` before applying the next delta and emits when `tracker_changed && (had_unified_diff || unified_diff.is_some())`, passing `unified_diff.unwrap_or_default()`. A non-empty → empty transition therefore emits `turn/diff/updated` with `diff: ""`; app-server forwards that value unchanged. Codex also has explicit `net_zero_patch_emits_empty_turn_diff` and `invalidation_emits_empty_turn_diff` tests. The adapter already replaces its cached value on empty snapshots and tests that behavior. I added a source-specific code comment and clarified the regression-test name in 6f7e8bb; no lifecycle workaround is needed.

